PR #1, opened 2020-12-19
fix typo that caused provisioners not to load

SwampDragons:


plugin/set.go
```diff
@@ -113,7 +113,7 @@ func (i *Set) start(kind, name string) error {
 		err = server.RegisterBuilder(i.Builders[name])
 	case "post-processor":
 		err = server.RegisterPostProcessor(i.PostProcessors[name])
-	case "provisioners":
+	case "provisioner":
```
azr, line 116: Woops ! I wonder if we should make of these const/shared vars

azr (approved):

plugin/set.go
SwampDragons: I think that probably makes sense.

PR #2, opened 2021-01-01
fix: `Unable to use http_interface with Proxmox provider hashicorp/packer#10369`

blz-ea:
Fix: hashicorp/packer#10369


azr (approved):
LGTM

PR #4, opened 2021-01-04
add pkrplugin package that helps finding plugin binaries.

azr:


azr:
( will re-open when this project compiles )

PR #9, opened 2021-01-08
Add datasources to plugin set

sylviamoss:


version/version.go
```diff
@@ -12,6 +12,14 @@ import (
 // The git commit that was compiled. This will be filled in by the compiler.
 var GitCommit string
 
+// Package version helps plugin creators set and track the sdk version using
+const Version = "0.0.6"
```
azr, line 16: nice !

plugin/set.go
```diff
@@ -41,19 +40,21 @@ type SetDescription struct {
...
 }
 
 ////
 // Setup
 ////
 
 func NewSet() *Set {
+	sdkVersion := pluginVersion.InitializePluginVersion(pluginVersion.Version, pluginVersion.VersionPrerelease)
 	return &Set{
-		version:        version.String(),
```
azr, line 52: oh right, we don't need that one there !

azr (approved):
nice one !

PR #10, opened 2021-01-08
Run tests on circleci, and fix busted makefile stuff from packer export

SwampDragons:
This Pr gets all but one set of tests running -- the download tests don't work on windows, so I've disabled them. I've traced this to a line ending issue, and will chase it down further on Monday, but this needs to be merged so that the .gitattributes can be read and hopefully properly check out line endings as a result.


multistep/commonsteps/step_download.go
```diff
@@ -65,6 +65,7 @@ func init() {
 }
 
 func (s *StepDownload) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	log.Printf("/////////// NEW RUN CASE/////////////// ")
```
sylviamoss, line 68: This seems to be a testing log 🤔

sylviamoss (approved):
Nice!! 
Some documentation was generated. Was this expected?

Makefile
```diff
@@ -14,7 +14,7 @@ install-gen-deps: ## Install dependencies for code generation
 	# out code dependencies; so a go mod tidy will remove them again. `go
 	# install` seems to install the last tagged version and we want to install
 	# master.
-	@(cd $(TEMPDIR) && GO111MODULE=on go get github.com/mna/@master)
+	@(cd $(TEMPDIR) && GO111MODULE=on go get github.com/mna/pigeon@master)
 	@(cd $(TEMPDIR) && GO111MODULE=on go get github.com/alvaroloes/enumer@master)
```
azr, line 18: Hello there, as per https://github.com/golang/go/issues/25922, another thing we _could_ do here to not have to run these steps is having a `tools.go` file that could look like so:

```go
// +build tools

package tools

import (
  _ "github.com/mna/pigeon"
  _ "github.com/alvaroloes/enumer"
)
```

This way these tools will be checked in the code for us and versioned by go.
Another cool thing we can do is change the go generate calls with something like:

```
//go:generate go run golang.org/x/tools/cmd/stringer@v1.11.0 -type=Pill
```

multistep/commonsteps/step_download.go
azr, line 68: Q: Is this a test/log ?

azr (commented):
Howdy there, I had a few suggestions to not have to handle version from the Makefile. LGTM otherwise.

multistep/commonsteps/step_download.go
SwampDragons: yeah sorry I'll remove it

SwampDragons:
@sylviamoss yeah the generated docs are intentional. 
@azr cool ideas -- I'll take a look at doing that today.

Makefile
SwampDragons: looks like the latter suggestion doesn't actually work; it was an idea. I'm tinkering with trying to get a to tools file working now

SwampDragons: I spent 15 minutes tinkering with this and couldn't get it working so I reverted; if we end up having to manage more dependencies in the future, or if golang ends up actually implementing something along the lines of 

>//go:generate go run golang.org/x/tools/cmd/stringer@v1.11.0 -type=Pill 

(which doesn't currently work, it was just an idea in that thread), then we can revisit it.

azr: Ah right bummer, sorry ! If the projects had more tags the header version could have worked, but it's not the case.

PR #11, opened 2021-01-11
Add hcl2 helper funcs

sylviamoss:
These two functions will be very useful for datasource creators to easily transform their output to a propper `cty.Value`.


SwampDragons:
I did a quick search of the linked PR and it looks like they're used in the hcl2upgrade command but not actually needed by the plugin itself, is that right?  Can I hear your reasoning for moving it?

sylviamoss:
About the `HCL2ValueFromConfigValue`, my thought was that if one wants to return a single config value in a Datasource execution, e.g. a map, then they can use this function to get the equivalent cty.Value. 
The other function is the one I created to help transform a struct output into a cty.Value. I can see both being very useful to someone implementing a Datasource.

hcl2/doc.go
```diff
@@ -0,0 +1,5 @@
+/*
+Package hcl2 provides helper functions for decoding interfaces of a single config value or
+a whole config struct into a cty.Value.
```
azr, line 3: Here, I'd avoid naming this pkg `hcl2`, to maybe name it to something that does not exist, also in the header we could say :

// Package ... provides helper functions for passing or getting hcl2 types to and from a Packer plugin.

Also, `hcl2.HCL2ValueFromConfigValue` and `hcl2.HCL2ValueFromConfig` will stutter.

In general, I'd advise on maybe thinking about the way the godocs api will look first here because that's what sdk using people will check out first.

sylviamoss: is hcl2shim a good name? the same we use in the packer core 

> Also, hcl2.HCL2ValueFromConfigValue and hcl2.HCL2ValueFromConfig will stutter.
What do you mean by that? 🤔

azr: Yes ! `hcl2shim` sounds like a good name ! Also, before seeing your message, after some thinking and because hcl is specific to each projects (terraform/consul/nomad/etc.) I though of maybe naming it `pkrhcl2helper` because its a helper for hcl2 that is specific to packer, I really like it but the name is kinda long I'd like to let you decide there.

Stutter means that the name of the function contains/repeats the name of the package, ex `hcl2.HCL2ValueFromConfig` and we could change it to: `hcl2.ValueFromConfig`, I reuse the wording from the go community on good practices.

sylviamoss: Nice, didn't know the term. Thanks! 
It makes sense everything you're saying. I'll make the changes here!

hcl2helper/doc.go
```diff
@@ -0,0 +1,4 @@
+/*
+Package hcl2helper provides helper functions for parsing or getting hcl2 types to and from a Packer plugin config.
+*/
+package hcl2helper
```
sylviamoss, line 4: I decided to leave the `pkr` suffix out of the name because being inside the packer sdk should be enough to separate logic from other hcl2 helpers from another tool. If one needs some discrepancy, they can always override the name when importing.

hcl2helper/values.go
```diff
@@ -0,0 +1,131 @@
...
+
+// UnknownVariableValue is a sentinel value that can be used
+// to denote that the value of a variable is unknown at this time.
+// RawConfig uses this information to build up data about
+// unknown keys.
+const UnknownVariableValue = "74D93920-ED26-11E3-AC10-0800200C9A66"
+
+// HCL2ValueFromConfigValue takes a value turns it into
+// a cty.Value so it can be used within, for example, an HCL2 EvalContext.
+func HCL2ValueFromConfigValue(v interface{}) cty.Value {
```
sylviamoss, line 22: Thought of changing this to `CTYValueFromConfigValue` but I didn't like it so I kept the HCL2 to keep it more clear.  `hcl2helper.HCL2ValueFromConfigValue` doesn't sound repeatedly like `hcl2.HCL2ValueFromConfigValue`

hcl2helper/doc.go
SwampDragons: I agree; this is a good name.

SwampDragons (approved):

PR #12, opened 2021-01-11
Move back Maps of Plugin in Packer core

azr:
Related: https://github.com/hashicorp/packer/pull/10466


SwampDragons:
I already left a comment and question on the core-side compliment to this PR -- we can chat about it on a zoom tomorrow.

azr:
I agree with you there, so I'm closing this PR then ! To leave the `MapOf...` types here **and** there.

PR #17, opened 2021-01-15
Add Release workflow to project - Testing DNM

nywilken:
This PR has been superseded by #19. This PR is being kept open for testing purposes.


azr (approved):

.circleci/config.yml
```diff
@@ -87,6 +87,13 @@ jobs:
     steps:
       - checkout
       - run: make generate-check
+  "sdk release":
+    executor: golang
+    working_directory: /go/src/github.com/hashicorp/packer-plugin-sdk
```
sylviamoss, line 92: Do we need to set this working_directory on every job? I notice not all of them set it.

sylviamoss (commented):
Nice!! 👏🏼  
I'm not approving yet because I noticed is a work in progress. I suppose you will still add `master` filters and fix the scripts with packer-plugin-sdk data. 
I had only a general question.

.circleci/config.yml
sylviamoss: I would guess that the job would be in the repo folder by default base on the ones that don't set this field.

nywilken: That's a good question. I don't suspect it is needed since we are using Go modules on for this project. I will adjust for this PR. Then open a new one to remove it from the other jobs once this is ready to go.

PR #20, opened 2021-01-20
Add NormalizeJsonString helper

sylviamoss:
Adding function that will be used by the next `amazon-secretsmanager-secret` data source. Also add a missing part to the data source acc test structure.

### DIFF
--- a/json/helper.go
+++ b/json/helper.go
@@ -38,3 +38,24 @@ func Unmarshal(data []byte, i interface{}) error {
 
 	return nil
 }
+
+// Takes a value containing JSON string and passes it through
+// the JSON parser to normalize it, returns either a parsing
+// error or normalized JSON string.
+func NormalizeJsonString(jsonString interface{}) (string, error) {
+	var j interface{}
+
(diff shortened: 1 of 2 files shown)


json/helper.go
```diff
@@ -38,3 +38,24 @@ func Unmarshal(data []byte, i interface{}) error {
...
+// the JSON parser to normalize it, returns either a parsing
+// error or normalized JSON string.
+func NormalizeJsonString(jsonString interface{}) (string, error) {
+	var j interface{}
+
+	if jsonString == nil || jsonString.(string) == "" {
+		return "", nil
+	}
+
+	s := jsonString.(string)
```
azr, line 52: This could be simplified a bit:

```suggestion
	s, ok := jsonString.(type)
	if !ok || s == "" {
	  return "", nil
	}
```

azr (approved):
Nice one ! I think it could be a good idea to add a tiny test ( I don't think it's mandatory )

nywilken (approved):
LGTM

sylviamoss:
No need for this PR anymore.

PR #28, opened 2021-01-25
Add generic plugin acc test

sylviamoss:
This generic plugin acceptance test is exactly the same used for data sources but with a generic name so that can be used for all plugins. The idea is to be config language agnostic since it will write a file with the given content and extension and run `packer build` against it. 

The result would be like the acceptance tests added to https://github.com/hashicorp/packer-plugin-scaffolding/pull/6 using this structure.

### DIFF
--- a/acctest/pluginacc.go
+++ b/acctest/pluginacc.go
@@ -10,13 +10,13 @@ import (
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


nywilken (approved):
LGTM :+1:

SwampDragons (approved):
looks good!

PR #30, opened 2021-02-04
Create dependabot.yml

nywilken:
Add dependabot to the repo to help keep go modules up to date. This is being done for tracking Packer and HCL pkgs. 
AWS package is being ignored because it is released every week.


azr (approved):
Great idea ! LGTM !

PR #31, opened 2021-02-06
Update ugorji/go/codec to v1.2.4

angdraug:
This is a continuation of hashicorp/packer#10146 that was submitted before the rpc code was moved from packer to packer-plugin-sdk.

When running Packer built with modern[0] version of ugorji using a plugin built from source and therefore with the 2015 version currently pinned in Packer's go.mod, I hit ugorji/go#269 when my plugin tries to [pipe into a remote "cat > '%s'"](https://github.com/angdraug/packer-builder-nspawn-debootstrap/blob/master/builder/communicator.go#L33):

>  Build 'base-20201021-5f90bdf5-490e-582b-fb1d-b5d4315f5eaa' errored after 25 seconds 772 milliseconds: msgpack decode error [pos 274]: invalid length of bytes for decoding time - expecting 4 or 8 or 12, got 15

[0] As in, newer than ugorji/go@5a66da2 from 2017, e.g. Packer as packaged in Debian.

The same pipe works fine when either running Packer built from source (with the same pin to old ugorji), or forcing the plugin to build with the modern ugorji:

```
replace github.com/hashicorp/packer => github.com/angdraug/packer v1.6.4-ugorji-go-v1.1.13
```

Bringing ugorji in Packer to the current version will leap over this and other incompatibilities that have built up in ugorji over the years, will make custom plugins work out of the box with the Debian build of Packer, and will allow plugin writers to use modules that depend on newer versions of ugorji.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/pkg/sftp v0.0.0-20160118190721-e84cc8c755ca
 	github.com/ryanuber/go-glob v1.0.0
 	github.com/stretchr/testify v1.6.1
-	github.com/ugorji/go v0.0.0-20151218193438-646ae4a518c1
+	github.com/ugorji/go/codec v1.2.4
 	github.com/zclconf/go-cty v1.7.0
 	golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9
 	golang.org/x/mobile v0.0.0-20201208152944-da85bec010a2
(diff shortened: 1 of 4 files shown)


azr (approved):
Hello there, thanks for opening again 🙂 . This looks fine by me; I did some testing and some acceptance testing too.

I think we should merge this now before Packer 1.7.0 is released, since we bumped the API MAJOR version.
We should also probably do much more testing; and some reading of the diff for the lib.
This dependency could become a difficult tech debt; and even if we introduce a bug by merging this; the lib seems well maintained so that we could PR our way out of a bug.

I think that the whole Packer team needs to agree with this though !

For context: the codec codec is what marshals a go struct into a string and back. In core to plugins communications, it's used there:

https://github.com/hashicorp/packer-plugin-sdk/blob/f0882e2e5bc73ec656ce4a22badc5180231507a2/rpc/client.go#L40-L56

and there:

https://github.com/hashicorp/packer-plugin-sdk/blob/f0882e2e5bc73ec656ce4a22badc5180231507a2/rpc/server.go#L143-L160

azr (approved):
Dorry, a Github 500 caused a dupe review, and I can't delete this one.

azr:
Got confirmation during our weekly meeting !

PR #40, opened 2021-02-18
Update links to documentation

gordonbondon:
Docs have been recently updated so previous links in readme went to 404


nywilken (approved):
Good catch. Thanks for making the fix.

PR #41, opened 2021-02-18
add more context to packer vault func error message

SwampDragons:
More helpful logging to direct user to packer vault docs
Closes https://github.com/hashicorp/packer/issues/10331


nywilken (approved):
Nice

PR #42, opened 2021-02-19
[wip] Add custom MapOfInterfaceToMapOfCTY decoder

sylviamoss:
Changes needed to https://github.com/hashicorp/packer/pull/10655
This will allow config as `map[string]cty.Value` which is the valid option for when `map[string]interface{}` is needed.


sylviamoss:
No longer necessary.

PR #43, opened 2021-03-19
`HTTPConfig `: Add http_content to be served

azr:
This will allow to

* [x] feature
* [x] docs
* [x] tests ; I also added some tests for the http config struct.

### DIFF
--- a/multistep/commonsteps/http_config.go
+++ b/multistep/commonsteps/http_config.go
@@ -26,9 +26,11 @@ type HTTPConfig struct {
 	// Key/Values to serve using an HTTP server. http_content works like and
 	// conflicts with http_directory The keys represent the paths and the values
 	// contents. This is useful for hosting kickstart files and so on. By
-	// default this is emptyu, which means no HTTP server will be started. The
+	// default this is empty, which means no HTTP server will be started. The
 	// address and port of the HTTP server will be available as variables in
-	// `boot_command`. This is covered in more detail below.
+	// `boot_command`. This is covered in more detail below. Example: Setting
+	// `"foo/bar"="baz", will allow you to http get on
+	// `http://{http_ip}:{http_port}/foo/bar`.
 	HTTPContent map[string]string `mapstructure:"http_content"`
 	// These are the minimum and maximum port to use for the HTTP server
 	// started to serve the `http_directory`. Because Packer often runs in
--- a/multistep/commonsteps/http_config.go
+++ b/multistep/commonsteps/http_config.go
@@ -23,6 +23,13 @@ type HTTPConfig struct {
 	// started. The address and port of the HTTP server will be available as
 	// variables in `boot_command`. This is covered in more detail below.
 	HTTPDir string `mapstructure:"http_directory"`
+	// Key/Values to serve using an HTTP server. http_content works like and
+	// conflicts with http_directory The keys represent the paths and the values
+	// contents. This is useful for hosting kickstart files and so on. By
+	// default this is emptyu, which means no HTTP server will be started. The
+	// address and port of the HTTP server will be available as variables in
+	// `boot_command`. This is covered in more detail below.
+	HTTPContent map[string]string `mapstructure:"http_content"`
 	// These are the minimum and maximum port to use for the HTTP server
 	// started to serve the `http_directory`. Because Packer often runs in
 	// parallel, Packer will choose a randomly available port in this range to
@@ -64,6 +71,11 @@ func (c *HTTPConfig) Prepare(ctx *interpolate.Context) []error {
 	if c.HTTPPortMin > c.HTTPPortMax {
 		errs = append(errs,
 			errors.New("http_port_min must be less than http_port_max"))
+	}
+
+	if len(c.HTTPContent) > 0 && len(c.HTTPDir) > 0 {
+		errs = append(errs,
+			errors.New("http_content cannot be used in conjunction with http_dir, consider using the file function"))
 	}
 
 	return errs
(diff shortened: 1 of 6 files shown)


multistep/commonsteps/http_config.go
```diff
@@ -23,6 +23,15 @@ type HTTPConfig struct {
...
 	HTTPDir string `mapstructure:"http_directory"`
+	// Key/Values to serve using an HTTP server. http_content works like and
+	// conflicts with http_directory The keys represent the paths and the values
+	// contents. This is useful for hosting kickstart files and so on. By
+	// default this is empty, which means no HTTP server will be started. The
+	// address and port of the HTTP server will be available as variables in
+	// `boot_command`. This is covered in more detail below. Example: Setting
+	// `"foo/bar"="baz", will allow you to http get on
+	// `http://{http_ip}:{http_port}/foo/bar`.
+	HTTPContent map[string]string `mapstructure:"http_content"`
```
sylviamoss, line 34: This is not a backward incompatibility but it is something that will require all builder components using http_config to update their hcl2spec files in order to be compatible with this feature.
If I'm right, I have two questions: 
- Are you planning to write a blog post for this?  
- Should this be officially released on 1.8.0?

multistep/commonsteps/step_http_server.go
```diff
@@ -22,23 +23,67 @@ import (
...
+		return
+	}
+
+	if _, err := w.Write([]byte(content)); err != nil {
+		// log err in case the file couldn't be 100% transferred for example.
+		log.Printf("http_content serve error: %v", err)
+	}
+}
+
+func (s *StepHTTPServer) Handler() http.Handler {
```
sylviamoss, line 61: This one up close to the StepHTTPServer helps understand it faster.

sylviamoss (commented):
Is this done? I made you some questions.
I haven't tested but the code looks good so far.

multistep/commonsteps/step_http_server.go
azr: good one !

multistep/commonsteps/http_config.go
azr: Good questions!
> Should this be officially released on 1.8.0?

The earlier we merge this one, the less manual updates we will have to do in plugins in order to support this. So my guess would be to merge this one now-ish.


> Are you planning to write a blog post for this?

🤔  I wasn't planning on doing that ! But you are right !
This definitely could use a new guide or help section too.

sylviamoss: Nice! We just need to remember to run generate on Packer when this one is merged. So that 1.7.1 will be compatible with these changes.

multistep/commonsteps/step_http_server.go
```diff
@@ -22,23 +26,76 @@ import (
...
+	if !found {
+		paths := make([]string, 0, len(s))
+		for k := range s {
+			paths = append(paths, k)
+		}
+		sort.Strings(paths)
+		err := fmt.Sprintf("%s not found.", path)
+		if sug := didyoumean.NameSuggestion(path, paths); sug != "" {
+			err += fmt.Sprintf("Did you mean %q?", sug)
+		}
```
azr, line 69: This makes lives easier <3 !

azr: ( even saved me a few times in tests mode already )

azr:
This is ready for review 🙂

SwampDragons:
I pulled this to try and test and I can't get it working; on at least on the QEMU builder it is trying to read from a port 0 which seems wrong. Can you give a more complete example?

multistep/commonsteps/http_config.go
```diff
@@ -66,5 +75,10 @@ func (c *HTTPConfig) Prepare(ctx *interpolate.Context) []error {
 			errors.New("http_port_min must be less than http_port_max"))
 	}
 
+	if len(c.HTTPContent) > 0 && len(c.HTTPDir) > 0 {
+		errs = append(errs,
+			errors.New("http_content cannot be used in conjunction with http_dir, consider using the file function"))
```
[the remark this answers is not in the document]
azr:
```suggestion
			errors.New("http_content cannot be used in conjunction with http_dir. Consider using the file function to load file in memory and serve them with http_content: https://www.packer.io/docs/templates/hcl_templates/functions/file/file"))
```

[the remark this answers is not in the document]
azr: Thanks !!! ( this is applied, tell me if something's up ! )

multistep/commonsteps/step_http_server.go
```diff
@@ -70,6 +126,8 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 func (s *StepHTTPServer) Cleanup(multistep.StateBag) {
 	if s.l != nil {
 		// Close the listener so that the HTTP server stops
-		s.l.Close()
+		if err := s.l.Close(); err != nil {
+			log.Printf("Failed closing http server on port %d: %v", s.l.Port, err)
```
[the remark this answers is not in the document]
azr: Good question, this error should be temporary and won't prevent Packer from working. We check if ports are already open for new calls to `Listen`. Finally, when Packer terminates: the connection will be auto closed on most systems. Though on some system it could take some time and cause a FS leak if it happened A LOT of times. Yup this should be at least be shown in the UI.

sylviamoss (approved):
LGTM!
Tested with virtualbox-iso building and alpine image and it worked good! 👍🏼

PR #44, opened 2021-03-23
[WIP] Http server addr

SwampDragons:
We extracted the SDK out from under https://github.com/hashicorp/packer/pull/10273; this applies the patch from that PR, preserving @redgryphon 's  commit credit. 

Note: this isn't ready for merge. We have to think carefully about how http_ip is implemented in different builders. It may be safer to only write here, never read -- and update various builders' implementations to make sure that they don't try to set the http_ip to state before this point (looking at you, qemu builder)

Closes https://github.com/hashicorp/packer/issues/9973

### DIFF
--- a/multistep/commonsteps/step_http_server.go
+++ b/multistep/commonsteps/step_http_server.go
@@ -111,13 +111,20 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Starting HTTP server on port %d", s.l.Port))
+	httpAddr, present = state.GetOK("http_ip")
+	if present {
+		if httpAddr.(string) == "" {
+			httpAddr = s.HTTPAddress
+		}
+        }
+	ui.Say(fmt.Sprintf("Starting HTTP server on address %s and port %d", httpAddr, s.l.Port))
 
 	// Start the HTTP server and run it in the background
 	server := &http.Server{Addr: "", Handler: s.Handler()}
 	go server.Serve(s.l)
 
-	// Save the address into the state so it can be accessed in the future
+	// Save the address and the port into the state so they can be accessed in the future
+	state.Put("http_ip", httpAddr)
 	state.Put("http_port", s.l.Port)
 
 	return multistep.ActionContinue
--- a/multistep/commonsteps/step_http_server.go
+++ b/multistep/commonsteps/step_http_server.go
@@ -97,6 +97,7 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 
 	// Find an available TCP port for our HTTP server
 	var err error
+	var httpAddr string
 	s.l, err = net.ListenRangeConfig{
 		Min:     s.HTTPPortMin,
 		Max:     s.HTTPPortMax,
@@ -111,12 +112,12 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	httpAddr, present = state.GetOK("http_ip")
-	if present {
-		if httpAddr.(string) == "" {
-			httpAddr = s.HTTPAddress
-		}
-        }
+	httpAddr = s.HTTPAddress
+	httpAddrIface, present := state.GetOk("http_ip")
+	if present && httpAddrIface.(string) != "" {
+		httpAddr = httpAddrIface.(string)
+	}
+
 	ui.Say(fmt.Sprintf("Starting HTTP server on address %s and port %d", httpAddr, s.l.Port))
 
 	// Start the HTTP server and run it in the background


multistep/commonsteps/step_http_server.go
```diff
@@ -54,14 +54,21 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Starting HTTP server on port %d", s.l.Port))
+	httpAddr, present = state.GetOK("http_ip")
```
nywilken, line 57:
```suggestion
	httpAddr, present = state.GetOk("http_ip")
```

multistep/commonsteps/step_http_server.go
```diff
@@ -54,14 +54,21 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Starting HTTP server on port %d", s.l.Port))
+	httpAddr, present = state.GetOk("http_ip")
```
nywilken, line 57:
```suggestion
	httpAddr, present := state.GetOk("http_ip")
```

multistep/commonsteps/step_http_server.go
```diff
@@ -54,14 +54,21 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
...
+	if present {
+		if httpAddr.(string) == "" {
+			httpAddr = s.HTTPAddress
+		}
+        }
```
nywilken, line 62: nit: on the nested if statements

```suggestion
	if present && httpAddr.(string) == "" {
			httpAddr = s.HTTPAddress
        }
```

nywilken (approved):
One small fix and a nit. But other wise good. Merge when ready.

SwampDragons:
Note: this isn't ready for merge. We have to think carefully about how http_ip is implemented in different builders. It may be safer to _only write_ here, never read -- and update various builders' implementations to make sure that they don't try to set the http_ip to state before this point (looking at you, qemu builder)

multistep/commonsteps/step_http_server.go
```diff
@@ -111,13 +112,20 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Starting HTTP server on port %d", s.l.Port))
+	httpAddr = s.HTTPAddress
+	httpAddrIface, present := state.GetOk("http_ip")
+	if present && httpAddrIface.(string) != "" {
+		httpAddr = httpAddrIface.(string)
+	}
```
sylviamoss, line 119: To cast `string` only once
```suggestion
	httpAddrIface, present := state.Get("http_ip").(string)
	if present && httpAddrIface != "" {
		httpAddr = httpAddrIface
	}
```

sylviamoss (commented):
What's the best test case to test this one? 
I ran a simple virtualbox build against these changes using the new `http_content` and it worked fine just like before.

multistep/commonsteps/step_http_server.go
azr: You can even simplify it to :

```suggestion
	httpAddrIface, _ := state.Get("http_ip").(string)
	if httpAddrIface != "" {
		httpAddr = httpAddrIface
	}
```

multistep/commonsteps/step_http_server.go
```diff
@@ -111,13 +112,20 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
...
+
+	ui.Say(fmt.Sprintf("Starting HTTP server on address %s and port %d", httpAddr, s.l.Port))
 
 	// Start the HTTP server and run it in the background
 	server := &http.Server{Addr: "", Handler: s.Handler()}
```
azr, line 124: I think we need to pass addr here:

```suggestion
	server := &http.Server{Addr: httpAddr, Handler: s.Handler()}
```

azr (approved):
LGTM, I'd add a unit test maybe just to make sure

multistep/commonsteps/step_http_server.go
```diff
@@ -111,13 +112,20 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
...
 	go server.Serve(s.l)
 
-	// Save the address into the state so it can be accessed in the future
+	// Save the address and the port into the state so they can be accessed in the future
+	state.Put("http_ip", httpAddr)
```
azr, line 128: do you think it could be a good idea to use http_addr here
```suggestion
	state.Put("http_addr", httpAddr)
```
It's probably just sort of a convention but I've always said you listen on an address and you connect to an IP ( maybe there's more scripture on that; trying to look it up )

multistep/commonsteps/step_http_server.go
azr: Also addr is used in `Addr: s.HTTPAddress,`, I think we should that one there too

nywilken:
Converting to a draft for the time being, as there are some follows up to this PR that we need to work out.

nywilken:
@SwampDragons this is a bit stale and we haven't had a change to revisit it. I'm going to close for now and revisit when needed. Thanks for kicking this off.

PR #48, opened 2021-04-01
Bump golang image to 1.16

nywilken:


sylviamoss (approved):
The sdk is on 1.14 though. I'm wondering if that's a problem but I think is not. Our local env is on 1.16 and it ran well 👍🏼

SwampDragons:
we should bump it too.

nywilken:
Oh okay I will bump that as well. 🤞

PR #51, opened 2021-04-07
Remove packer core dependencies 

sylviamoss:
Removing all of the packer core dependencies here. 

Regarding the example, we already have documentation with good examples of serving a multi-component plugin.

I'm moving the `testing.go` and `testing_test.go` files to the packer core. This a temporary solution while we break out plugins and update their acc tests to use the framework from https://github.com/hashicorp/packer-plugin-sdk/blob/main/acctest/pluginacc.go


plugin/example/main.go
```diff
@@ -1,36 +0,0 @@
-// This is an example plugin. In packer 0.9.0 and up, core plugins are compiled
```
azr, line 1: Yup, I agree here this example is in the scaffolding !

azr (approved):
LGTM

SwampDragons (approved):

PR #54, opened 2021-04-14
Testutils

SwampDragons:
Add a new testutil for loading a manifest file so we can retrieve artifact ids in test check functions.

### DIFF
--- a/acctest/testutils/utils.go
+++ b/acctest/testutils/utils.go
@@ -2,7 +2,12 @@
 // acceptance testing.
 package testutils
 
-import "os"
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+)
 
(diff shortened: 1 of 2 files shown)


acctest/testutils/utils.go
```diff
@@ -18,3 +23,60 @@ func FileExists(filename string) bool {
 	}
 	return false
 }
+
+// Use the manifest to load information about artifact ID and region.
+// to properly use this, you should make sure the manifestfilepath is unique
+// for each test, and remember to clean up the manifest file when your build
+// is done!
```
sylviamoss, line 30: nice!!

sylviamoss (approved):
LGTM!! It would be nice to have a test just to make sure, but that can be added later as this is very simple. 👍🏼

PR #59, opened 2021-04-23
Update GPG usage

shore:
As part of rotating the GPG key used for signing products, we are also moving the GPG bits into a dedicated CircleCI context.  This PR renames the GPG env vars to match those from the new context and applies the context to the job that uses the key.

This was tested locally by modifying the `release.sh` script to not push commits, setting the environment variables to match my local keyring, running the script, and inspecting the produced commit with `git log -n 1 --show-signature`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,9 +23,9 @@ steps:
     run:
       name: import base64 GPG key and prime passphrase
       command: |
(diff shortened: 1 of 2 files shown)


azr (approved):
Awesome, thanks !

SwampDragons (approved):

PR #67, opened 2021-06-02
Capitalize some parts of the output

picatz:
👋 Hello! I noticed some parts of the Packer build output that seemed inconsistent or used a weird capitalization. For example: "ssh" instead of "SSH", or "rsa" instead of "RSA". This PR introduces some capitalization changes to the user-facing CLI output to fix those cases.

These capitalizations are arguably really not _necessary_. Maybe this could even break assumptions for users that would want to parse Packer build output? Not totally sure. But, I figured I'd open a PR to get feedback. 😄


SwampDragons (approved):

SwampDragons:
I like it. Thanks!

PR #68, opened 2021-06-20
Feature/xdg base dir 2

teddylear:
Setting the 'PACKER_CONFIG_DIR` and `PACKER_CONFIG_DIR` to use the default XDG spec for unix systems. Not sure of the best way of testing these changes as they are operating system specific for the paths. Because of this I removed tests for cache directory as they changed depending on the operating system. If you have any advice for how to add tests for these please let me know!

Addresses: [hashicorp/packer#9007](https://github.com/hashicorp/packer/issues/9007)


packer/cache_config_unix.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func getDefaultCacheDir() string {
+	var defaultConfigFileDir string
+
+	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
```
ogarcia, line 13: This is bad, not is `XDG_CONFIG_HOME` is `XDG_CACHE_HOME`

ogarcia (changes requested):

packer/cache_config_unix.go
teddylear: Ah let me fix.

teddylear: I have pushed update to fix this

SwampDragons:
Hi, thanks for opening this. I haven't had a chance to test yet, but this appears like it would be backwards-breaking in that new versions of Packer will look in a new place and error if the XDG_CACHE_HOME env var exists, but the config file isn't there. 

I think to preserve backwards compatability it would be better to first check for the config in the XDG_CACHE_HOME and then fall back to checking the old path if nothing is discovered in XDG_CACHE_HOME.

teddylear:
@SwampDragons I updated PR to check for XDG_CONFIG_HOME env car and for packer configuration there before defaulting back to old location. I'm assuming for the cache it does not matter the location as it's a cache. Please let me know if there's anything I should change and if there is a way to automate the testing of this.

ogarcia:
@teddylear bad update, `XDG_CONFIG_HOME` does not have to be configured or exist as environment variable. The behavior must be:
1. If `XDG_CONFIG_HOME` exists as environment variable use it
2. If not use default: `$HOME/.config`

I think that you must check if old config dir exists and in this case use it, if not, use XDG.

teddylear:
@SwampDragons I just want to confirm the behavior that we want here. Do we want the behavior listed on the issue that @ogarcia is suggesting with: 
   1. If `XDG_CONFIG_HOME` exists as environment variable use it
   2. If not use default: `$HOME/.config`

Or the behavior I believe your last comment suggests which is: 
1. Check if `XDG_CONFIG_HOME` and if there is configuration there then use it
2. else use the default location we have already today. 

I'm assuming for the cache we do not mind where this is placed as it's a cache. Please let me know how to proceed.

ogarcia:
@teddylear you forgot one step:
1. Check if default location that we have today exists, if yes, use it (backwards compatibility).
2. Else If `XDG_CONFIG_HOME` exists as environment variable use it
3. Else use default: `$HOME/.config`

With this implementation, the new users have XDG, and old ones can delete default location or move current config to XDG to have it.

SwampDragons:
I think either mine as you restated it, or ogarcia's work. I'd originally skipped the "check home, then xdg" because I figure if users have a packer config in the xdg-set location it's intentional and should be used, but I can see the benefit of keeping pure backwards compat.

ogarcia:
@SwampDragons the main reason of perform check before is that all Linux/MacOS/Derivatives users are using XDG because is a standard (even if the `XDG_CONFIG_HOME` environment variable is not explicity declared).

In the other hand, using your approach for a user who was not configuration, if the app need to create something in its config directory will use default dir instead of `XDG_CONFIG_HOME` because `XDG_CONFIG_HOME` is empty for these user.

teddylear:
I have updated with @ogarcia suggestion. Please let me know if there is anything to update or change.

ogarcia:
@teddylear mutch better! But you must fix `XDG_CACHE_HOME` because now it have the logic:
1. If `XDG_CACHE_HOME` exists it will be used
2. Else, use a default `packer_cache` directory

And must be:
1. If `XDG_CACHE_HOME` exists as environment variable use it
2. If not use default: `$HOME/.cache`

Or use same logic that `XDG_CONFIG_HOME` if you want backwards compatibility, but is a cache and I think that is not needed in this case. 😉

teddylear:
@ogarcia I believe I have updated with the desired logic.

packer/cache_config_unix.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func getDefaultCacheDir() string {
+	var defaultConfigFileDir string
+
+	if xdgCacheHome := os.Getenv("XDG_CACHE_HOME"); xdgCacheHome != "" {
+		defaultConfigFileDir = filepath.Join(xdgCacheHome, "packer")
+	} else {
+		homeDir := os.Getenv("HOME")
+		defaultConfigFileDir = filepath.Join(homeDir, ".cache")
```
ogarcia, line 17: You miss add _packer_ here.
```go
defaultConfigFileDir = filepath.Join(homeDir, ".cache", "packer")
```

teddylear: Updated for cache and config.

pathing/config_file_unix.go
```diff
@@ -2,7 +2,42 @@
...
+	}
+	if cd := os.Getenv("PACKER_CONFIG_DIR"); cd != "" {
+		log.Printf("Detected config directory from env var: %s", cd)
+		dir = filepath.Join(cd, defaultConfigDir)
+	} else if hasDefaultConfigFileLocation(homedir) {
+		dir = filepath.Join(homedir, defaultConfigDir)
+		log.Printf("Old default config directory found: %s", dir)
+	} else if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
+		log.Printf("Detected xdg config directory from env var: %s", xdgConfigHome)
+		dir = xdgConfigHome
```
ogarcia, line 30: Here is miss the _packer_ too
```go
dir = filepath.Join(xdgConfigHome, "packer")
````

Comment (teddylear) [reply pathing/config_file_unix.go]: Updated

**Comment by ogarcia** [comment]
@teddylear Great work! IMHO it's ready to merge :wink:

**Comment by teddylear** [comment]
@SwampDragons This is ready to review, please let me know if you have any ideas around automating tests for this.

**Comment by SwampDragons** [comment]
sorry, I'm slammed. I'll try to take a look soon.

File: packer/cache.go [hunk 4]
```diff
@@ -5,16 +5,12 @@ import (
...
 // CachePath returns an absolute path to a cache file or directory
 //
-// When the directory is not absolute, CachePath will try to get
-// current working directory to be able to return a full path.
-// CachePath tries to create the resulting path if it doesn't exist.
-//
-// CachePath can error in case it cannot find the cwd.
+// When the directory is not absolute, CachePath will try to make a
+// a cache depending on the operating system.
 //
```
azr, line 12: 👋🏼 
Could you please update the examples down here in this list depending on the OS ? That would make it easier for me. 🙂

packer/cache.go
```diff
@@ -25,11 +21,15 @@ func CachePath(paths ...string) (path string, err error) {
...
 		cacheDir = cd
 	}
 
 	paths = append([]string{cacheDir}, paths...)
-	return filepath.Abs(filepath.Join(paths...))
+	result, err := filepath.Abs(filepath.Join(paths...))
+	if err != nil {
+		return "", err
+	}
+	return result, err
```
azr, line 34: Why not returning as is ?

packer/cache_test.go
```diff
@@ -1,52 +0,0 @@
-package packer
-
-import (
-	"os"
-	"path/filepath"
-	"testing"
-)
-
-func TestCachePath(t *testing.T) {
```
azr, line 9: Why removing the tests ?

pathing/config_file_unix.go
```diff
@@ -2,7 +2,42 @@
...
 	defaultConfigFile = ".packerconfig"
 	defaultConfigDir  = ".packer.d"
 )
+
+func configDir() (path string, err error) {
+	var dir string
+	homedir := os.Getenv("HOME")
+	if homedir == "" {
+		return "", err
+	}
```
azr, line 21: Since homedir is only being used for the `hasDefaultConfigFileLocation` and the last else cases, do you think we should `return ""` then ? I think this should probably error, or log something, too.

pathing/config_file_windows.go
```diff
@@ -2,7 +2,29 @@
...
 	defaultConfigFile = "packer.config"
 	defaultConfigDir  = "packer.d"
 )
+
+func configDir() (path string, err error) {
+	var dir string
+	if cd := os.Getenv("PACKER_CONFIG_DIR"); cd != "" {
+		log.Printf("Detected config directory from env var: %s", cd)
+		dir = filepath.Join(cd, defaultConfigDir)
+	} else {
```
azr, line 21: I'd make of this one a guard close:

```suggestion
	if cd := os.Getenv("PACKER_CONFIG_DIR"); cd != "" {
		log.Printf("Detected config directory from env var: %s", cd)
		dir = filepath.Join(cd, defaultConfigDir)
		return dir, nil
	}
```

packer/cache_config_unix.go
```diff
@@ -0,0 +1,21 @@
...
+	"os"
+	"path/filepath"
+)
+
+func getDefaultCacheDir() string {
+	var defaultConfigFileDir string
+
+	if xdgCacheHome := os.Getenv("XDG_CACHE_HOME"); xdgCacheHome != "" {
+		defaultConfigFileDir = filepath.Join(xdgCacheHome, "packer")
+	} else {
```
azr, line 15: NIT: same here:
```suggestion
	if xdgCacheHome := os.Getenv("XDG_CACHE_HOME"); xdgCacheHome != "" {
		return filepath.Join(xdgCacheHome, "packer")
	}
```
(this won't compile)

packer/cache_test.go
teddylear: As mentioned above, I'm having issues with creating tests with each path being different depending on the OS. If you have any examples of how to do this I would be glad to update as I tried to update the tests in place but hit a wall making them work for without needing to know the OS.

azr: Because the behaviors differ between OS, you will have to create test that test things under unix and under windows. One way to fix this would be to check using `runtime.GOOS == "windows"`, may be some table tests could be generated depending on the os type ?

packer/cache.go
teddylear: Updated.

packer/cache.go
teddylear: Fixed back what it was

packer/cache_test.go
teddylear: Updated to make different test files depending on OS similar to the OS specific cache and config files.

pathing/config_file_unix.go
teddylear: Updated with an error and added test case for error.

teddylear:
@azr Addressed previous comments. Please let me know if there is anything else to updates.

packer/cache_config_unix_test.go
```diff
@@ -0,0 +1,108 @@
...
+	os.Setenv("PACKER_CACHE_DIR", "")
+	defer func() {
+		os.Setenv("PACKER_CACHE_DIR", packerCacheDir)
+	}()
+
+	xdgCacheHomeDir := os.Getenv("XDG_CACHE_HOME")
+	os.Setenv("XDG_CACHE_HOME", "")
+	defer func() {
+		os.Setenv("XDG_CACHE_HOME", xdgCacheHomeDir)
+	}()
```
azr, line 36: Super nice !

azr: Do you think it could be a good idea to do something similar with `$HOME` ?

packer/cache_config_unix_test.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+			resetTestEnv()
+		})
+	}
+}
+
+func resetTestEnv() {
+	os.Setenv("PACKER_CACHE_DIR", "")
+	os.Setenv("XDG_CACHE_HOME", "")
+}
```
azr, line 108: Is it a leftover ?
```suggestion
```

packer/cache.go
azr: Super awesome, thanks.

pathing/config_file_unix_test.go
```diff
@@ -0,0 +1,141 @@
...
+			resetTestEnv()
+		})
+	}
+}
+
+func resetTestEnv() {
+	os.Setenv("PACKER_CONFIG_DIR", "")
+	os.Setenv("XDG_CONFIG_HOME", "")
+	os.Setenv("HOME", "")
+}
```
azr, line 141: Is this also a leftover ?
```suggestion
```

pathing/config_file_windows_test.go
```diff
@@ -0,0 +1,79 @@
...
+					tt.want)
+			}
+			resetTestEnv()
+		})
+	}
+}
+
+func resetTestEnv() {
+	os.Setenv("PACKER_CONFIG_DIR", "")
+}
```
azr, line 79: Is this also a leftover ?
```suggestion
```

packer/cache.go
```diff
@@ -5,27 +5,29 @@ import (
...
+// For Windows:
 //   PACKER_CACHE_DIR=""            CacheDir() => "./packer_cache/
 //   PACKER_CACHE_DIR=""            CacheDir("foo") => "./packer_cache/foo
 //   PACKER_CACHE_DIR="bar"         CacheDir("foo") => "./bar/foo
 //   PACKER_CACHE_DIR="/home/there" CacheDir("foo", "bar") => "/home/there/foo/bar
+// For Unix:
+//   PACKER_CACHE_DIR="",            XDG_CONFIG_HOME="", Default_config CacheDir() => "$HOME/cache/packer"
+//   PACKER_CACHE_DIR="",            XDG_CONFIG_HOME="", Default_config CacheDir("foo") => "$HOME/cache/packer/foo"
+//   PACKER_CACHE_DIR="bar",         XDG_CONFIG_HOME="", Default_config CacheDir("foo") => "./bar/foo
+//   PACKER_CACHE_DIR="/home/there", XDG_CONFIG_HOME="", Default_config CacheDir("foo", "bar") => "/home/there/foo/bar
```
azr, line 24: I like these new examples !
What will `Default_config` stand for here ?
Do you think we should add something for `XDG_CONFIG_HOME` too ?
```suggestion
// For Unix:
//   PACKER_CACHE_DIR="",            XDG_CONFIG_HOME="", Default_config CacheDir() => "$HOME/cache/packer"
//   PACKER_CACHE_DIR="",            XDG_CONFIG_HOME="", Default_config CacheDir("foo") => "$HOME/cache/packer/foo"
//   PACKER_CACHE_DIR="bar",         XDG_CONFIG_HOME="", Default_config CacheDir("foo") => "./bar/foo
//   PACKER_CACHE_DIR="/home/there", XDG_CONFIG_HOME="", Default_config CacheDir("foo", "bar") => "/home/there/foo/bar
//   PACKER_CACHE_DIR="", XDG_CONFIG_HOME="/home/there", Default_config CacheDir("foo", "bar") => "/home/there/foo/bar
```

azr:
Super awesome, thanks for bearing with us @teddylear, I just want to say that this LGTM in terms of code and I just had a question/suggestion on docs.

packer/cache_test.go
azr: Super good ! Thanks.

pathing/config_file_unix.go
azr: Dope !

SwampDragons:
Thanks for this!

PR #76, opened 2021-08-24
Add registry image pkg

nywilken:
In order to assist in the creation of HCP Packer Registry Image metadata
for various Artifacts. A new package is being added to assist in the
  creation of a registryimage.Image from an existing Artifact.

The FromArtifact function provides a simple approach of extracting the
required bits from an Artifact needed by the HCP Packer registry. The
default values, which can be overwritten by the use of a
ArtifactOverrideFunc , were selected to provide some standard defaults
that are set to be available for most plugins were bundled
with Packer prior to v1.7.0. When those defaults are not viable the
package provider override funcs that can be used by the consumer for
setting the appropriate values.

For those  plugins that generate more than one registryimage.Image per
Artifact one can construct a []registryimage.Image and return that
information as the response of Packer core calling
artifact.State(registryimage.ArtifactStateURI).

When reviewing, I recommend checking out the godoc to see the examples tests in action. 
```
> go install golang.org/x/tools/cmd/godoc@latest
> godoc -http=:8080
browse to http://localhost:8080/pkg/github.com/hashicorp/packer-plugin-sdk/packer/registry/image/
```


packer/registry/image/image.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// FromMappedData calls f sequentially for each key and value present in mappedData to create a []*Image
+// as the final return value. If there is an error in calling f, FromMappedData will stop processing mapped items
+// and result in a nil slice, with the said error.
+//
+// FromMappedData will make its best attempt to convert the input map into map[interface{}]interface{} before
+// calling f(k,v). The func f is responsible for type asserting the expected type for the key and value before
+// trying to create an Image from it.
+func FromMappedData(mappedData interface{}, f func(key, value interface{}) (*Image, error)) ([]*Image, error) {
```
nywilken, line 45: As an alternative to this implementation which would allow for a map of any type. I'm considering just supporting a mappedData input of map[string]string, as this seems to be the std. map being used across most plugins that return multi-region artifacts. 

It also removes the need for the consumer to do type assertions in the iteration function, which seems a bit complex.

azr (approved):
LGTM !

packer/registry/image/image.go
sylviamoss: Any specific reason you didn't go to `map[string]string` from the beginning?

sylviamoss (approved):
👍🏼  this looks great!

PR #77, opened 2021-08-30
import provisioner partials from Packer core

azr:
So that we can generate partial docs for inspec and salt provisioners


cmd/packer-sdc/internal/renderdocs/docs-partials/packer-plugin-sdk/provisioners/community_provisioners.mdx
```diff
@@ -0,0 +1,9 @@
+### Community Provisioners
```
sylviamoss, line 1: do we want this one here too?

sylviamoss (approved):
LGTM 👍🏼

cmd/packer-sdc/internal/renderdocs/docs-partials/packer-plugin-sdk/provisioners/community_provisioners.mdx
azr: Probably not, I just brought it with the folder, removing

PR #80, opened 2021-09-27
Implement callbacks for HTTP Server

### LINKED ISSUE #50: Whenever bootstrap files hosted in the http_directory are accessed, it should show in the logs
Here is a simple suggestion to save us lots of time in debugging.

In images like `virtualbox-iso`, one can host files using the `http_directory` parameter.

These files are are, for example, "install settings" used in debian and ubuntu.

On a good day, everything works as expected. 

On other days, it would be great if the internal HTTP server on packer showed some very basic access logs, so we could be sure our ISO image is hitting the HTTP server to download the files and this way narrow down whatever error we are having to "before" or "after" downloading the HTTP file.

flx5:
This pull request implements callbacks for requests to the http server.
That way plugins can inspect requests if required.
Additionally  a logging callback is added which should resolve #50.

Please let me know if any changes are required.

### DIFF
--- a/multistep/commonsteps/step_http_server.go
+++ b/multistep/commonsteps/step_http_server.go
@@ -16,13 +16,23 @@ import (
 )
 
 func HTTPServerFromHTTPConfig(cfg *HTTPConfig) *StepHTTPServer {
-	return &StepHTTPServer{
+	server := StepHTTPServer{
 		HTTPDir:     cfg.HTTPDir,
 		HTTPContent: cfg.HTTPContent,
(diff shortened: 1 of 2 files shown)


nywilken:
Hi @flx5 thanks again for making this change, and apologies for the delayed response here. Like the screenshot PR this is an interesting change and something that we think would be helpful to improving the debugging experience work we are envisioning for the future. 

Unfortunately at this time this is not a feature we have the cycles to take on. So I am going to close this PR but when ready I will gladly ping the thread and see if there is anything we can do to get this change merged and fully supported. Cheers!

PR #81, opened 2021-09-29
Add SourceImageID to PARtifact

SwampDragons:
This change will allow us to push the source image id to PAR more explicitly.


sylviamoss (approved):
Nice 👍🏼

packer/registry/image/fromartifact_example_test.go
```diff
@@ -55,7 +55,19 @@ func ExampleWithProvider() {
 	hcimage, _ := image.FromArtifact(a, image.WithProvider("happycloud"), image.WithRegion("west"))
 	fmt.Printf("%#v", *hcimage)
 	// Output:
-	// image.Image{ImageID:"service-id-123", ProviderName:"happycloud", ProviderRegion:"west", Labels:map[string]string{}}
+	// image.Image{ImageID:"service-id-123", ProviderName:"happycloud", ProviderRegion:"west", Labels:map[string]string{}, SourceImageID:""}
+}
+
+func ExampleWithSourceImageImageID() {
```
nywilken, line 61: This tests is failing because it is looking for a function with the name `WithSourceImageImageID`. I would update the example function to be like the following.
 
```suggestion
func ExampleWithSourceID() {
```

nywilken (approved):
This looks good to me ✅ - the failing tests is due to a naming issue. I added a suggestion which should fix it, I believe 🤔

nywilken:
Just a thought there is a Validate function for the PARtifact which is called in Packer core before adding the image to the build. Should SourceID be a required field? https://github.com/hashicorp/packer-plugin-sdk/blob/6505f23495e7ee59a8975d79033da1d8f2921615/packer/registry/image/image.go#L44

packer/registry/image/fromartifact_example_test.go
SwampDragons: Thanks!

SwampDragons:
Oh, good call. It probably should be required.

PR #82, opened 2021-10-01
Add `floppy_content` parameter in `FloppyConfig`

sestegra:
Add similar parameter as `http_content` and  `cd_content`, but for floppy.

This PR Implements a pre-requisite for supporting https://github.com/hashicorp/packer-plugin-vsphere/issues/110

### DIFF
--- a/multistep/commonsteps/floppy_config.go
+++ b/multistep/commonsteps/floppy_config.go
@@ -33,7 +33,24 @@ type FloppyConfig struct {
 	// characters (\\*, ?, and \[\]) are allowed. The maximum summary size of
 	// all files in the listed directories are the same as in `floppy_files`.
 	FloppyDirectories []string `mapstructure:"floppy_dirs"`
-	FloppyLabel       string   `mapstructure:"floppy_label"`
+	// Key/Values to add to the floppy disk. The keys represent the paths, and
+	// the values contents. It can be used alongside `floppy_files` or
+	// `floppy_dirs`, which is useful to add large files without loading them
+	// into memory. If any paths are specified by both, the contents in
+	// `floppy_content` will take precedence.
+	//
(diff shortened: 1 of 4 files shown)


multistep/commonsteps/floppy_config.go
```diff
@@ -33,7 +33,24 @@ type FloppyConfig struct {
...
+	// ```hcl
+	// floppy_files = ["vendor-data"]
+	// floppy_content = {
+	//   "meta-data" = jsonencode(local.instance_data)
+	//   "user-data" = templatefile("user-data", { packages = ["nginx"] })
+	// }
+	// floppy_label = "cidata"
+	// ```
+	FloppyContent map[string]string `mapstructure:"floppy_content"`
+	FloppyLabel   string            `mapstructure:"floppy_label"`
```
azr, line 53: `FloppyLabel` seems to not be used throughout the code. Do we need it then ? Is it that a builder would need it ? If so then I think that it should only be defined in the builder.

Also, I think it'd need a comment too 🙂 .

multistep/commonsteps/floppy_config.go
```diff
@@ -33,7 +33,24 @@ type FloppyConfig struct {
...
+	//
+	// ```hcl
+	// floppy_files = ["vendor-data"]
+	// floppy_content = {
+	//   "meta-data" = jsonencode(local.instance_data)
+	//   "user-data" = templatefile("user-data", { packages = ["nginx"] })
+	// }
+	// floppy_label = "cidata"
+	// ```
+	FloppyContent map[string]string `mapstructure:"floppy_content"`
```
azr, line 52: Super nice !

azr (approved):
This looks great, thanks !

PR #83, opened 2021-10-05
Add screenshot callback

flx5:
This creates a possibility for plugin maintainers to create screenshots on a failing connect.

Closes https://github.com/hashicorp/packer/issues/2990


nywilken:
Hi @flx5 thanks for making this change, and apologies for the delayed response here. The ability to support screenshots is interesting for sure and something that we might want to revisit at a later time. I think such a great feature requires documentation and some examples on how to use it. Especially since it is an advance configuration that will require plugin maintainers to setup their environment for supporting screenshots. 

Unfortunately at this time this is not a feature we have the cycles to take on. So I am going to close this PR but if things change near future I will gladly ping the thread and see if there is anything we can do to get this change merged and fully supported. Cheers!

PR #85, opened 2021-10-08
add pkr plugin check cmd from packer core

azr:
This is simplified to not pull in core as a dep. The drawback is that we don't actually start plugins to verify that it works with Packer and has a `ConfigSpec` func. We do start a binary though.

But I think this is fine since if a plugin implements describe then we can be pretty sure it uses the SDK's function and interfaces.

Related to https://github.com/hashicorp/packer/pull/11317

closes https://github.com/hashicorp/packer/issues/11309

### DIFF
--- a/cmd/packer-sdc/internal/plugincheck/cmd.go
+++ b/cmd/packer-sdc/internal/plugincheck/cmd.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


cmd/packer-sdc/internal/plugincheck/cmd.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil
+	}
+
+	if err := checkPluginName(pluginName); err != nil {
+		return err
+	}
+
+	output, err := exec.Command(pluginName, "describe").Output()
+	if err != nil {
+		return errors.Wrap(err, "failed to describe plugin")
```
nywilken, line 58: What do you think about adding an example to this error message? 
```suggestion
                s := fmt.Sprintf("failed to describe plugin using '%s describe'", pluginName)
		return errors.Wrap(err, s, )
```

PR #90, opened 2021-11-23
HCP: error handling & empty source image id

azr:


packer/registry/image/image.go
```diff
@@ -131,10 +123,6 @@ func WithProvider(name string) func(*Image) error {
 // used to override the ImageId for an existing Image.
 func WithID(id string) func(*Image) error {
 	return func(img *Image) error {
-		if img == nil {
-			return errors.New("no go on empty image")
-		}
```
azr, line 136: I think this was added to avoid a panic in case `FromMappedData` failed and returned a nil image. In this case, I think that this should panic, since the error was not checked. Also, this error is kind of weird.

packer/registry/image/image.go
```diff
@@ -50,10 +50,6 @@ func (i *Image) Validate() error {
 		return errors.New("error registry image does not contain a valid ProviderName")
 	}
 
-	if i.SourceImageID == "" {
-		return errors.New("error registry image does not contain a valid SourceImageID")
-	}
-
```
azr, line 56: Since this is optional, I think this should not be an error when this is not passed. Also, after taking a look around plugins, sometimes this was wrongly used, as in for the amazon plugin, the ami was used 🤔.

packer/registry/image/image.go
azr: (ditto throughout)

packer/registry/image/image.go
azr: I think it would be a good idea to have the image.`cloud_source_image_id` AND the build.`hcp_source_image_id` though.

packer/registry/image/image.go
azr: (I'm also super open to suggestions!)

packer/registry/image/image.go
sylviamoss: The API is supposed to get this AMI id and find the source image that has it as ID. So in HCP, this will be transformed into HCP-related data.

packer/registry/image/image.go
sylviamoss: The error is weird but panicking too. Maybe improving the error message is better 🤔  what do you think? Making it a friendly panic 😆

packer/registry/image/image.go
sylviamoss: But wait, in the amazon plugin, which AMI was used?

azr: Here: https://github.com/hashicorp/packer-plugin-amazon/blob/998d3209616d413eff1847c1bd7e5909b7b3c96c/builder/common/artifact.go#L167-L169

sylviamoss: Ahh ok, so that's right! The source AMI is set with the one queried from hcp, if using the data source. So the API will know who's the parent from it. 🤔

azr: Gotcha !

azr (commented):
In a 1:1 zoom with @nywilken, we are going to merge this one.

It is okay if SourceImageID is not set, and we can revisit that when we work on ancestry.

It's okay to panic when img is nil after an instantiation call failed.

PR #93, opened 2021-12-14
Update circleci image to use new cimg namespace with Go1.17

nywilken:


filelock/filelock.go
```diff
@@ -1,3 +1,4 @@
+//go:build !solaris
 // +build !solaris
```
azr, line 2: Should we only keep the newer//more readable version here ?

azr: Maybe not in case this gets built with go 1.16

nywilken: Oh interesting. I thought it was a direct replace of the build tag. Packer currently has both so maybe we can keep it consistent, and remove once 1.18 is out. What do you think?

nywilken: I'm merging as is for now. During our sync tomorrow we can make a call to add or remove here and in Packer if we don't want to wait until Go 1.18.

azr: yup sounds good to me !

azr:
Nice one !

PR #95, opened 2022-01-31
Append possibility to decompress archive with many files

LeConTesteur:
Hello,

I try to implement this issue hashicorp/packer-plugin-qemu#7.
The plugin use the step `stepDownload` for decompress archive, however the archive with many file is not supported.  
I append code for untar ova or other archive type into folder.
I update the version of `go-getter` for use `TarDecompressor`.

I append new tests.


nywilken:
Hi @LeConTesteur we ran into some issues with the untarring ova files in go-getter. So that change was reverted in [v2.0.2](https://github.com/hashicorp/go-getter/releases/tag/v2.0.2). This is not a change that we will be taking on in the SDK at this time, as it does change the behavior for all builders. 

If this change is specific to the QEMU builder I think the best approach would be to implement a StepDownload for the builder to not change the behavior for all plugins that use the shared StepDownload logic.

PR #98, opened 2022-02-15
all shells: Pass env vars through a key value store

azr:
* [x] feature
* [x] docs


nywilken (approved):
Nice add. She we consider deprecating the environment_vars field?

azr:
Ehm, good question, I would think that `environment_vars` sounds pretty widely used and should stay there forever.

PR #101, opened 2022-03-17
Provision SSHPublicKey while use custom SSH Private key

### LINKED ISSUE #100: Setup SSHPublicKey value while use custom SSHPrivateKeyFile
#### Description

Currently, [the code](https://github.com/hashicorp/packer-plugin-sdk/blob/622cc94659d8e673b0c1eae07b68c4e95c3b66be/communicator/step_ssh_keygen.go#L34), when using a user-specified SSH private key file, leaves the `SSHPublicKey` field explicitly empty. 
On the other hand when use  generated keypair value of SSHPublicKey [provisioned](https://github.com/hashicorp/packer-plugin-sdk/blob/622cc94659d8e673b0c1eae07b68c4e95c3b66be/communicator/step_ssh_keygen.go#L60).

#### Use Case(s)

Builder could use value of `SSHPublicKey` to correct provision instance when use custom private SSH key file.

GennadySpb:
Closes #100

### DIFF
--- a/communicator/sshkey/pubkey.go
+++ b/communicator/sshkey/pubkey.go
@@ -1,0 +1,16 @@
+package sshkey
+
+import (
+	"fmt"
+
+	"golang.org/x/crypto/ssh"
+)
+
+func PublicKeyFromPrivate(privateKeyBytes []byte) ([]byte, error) {
+	key, err := ssh.ParsePrivateKey(privateKeyBytes)
+	if err != nil {
+		return nil, fmt.Errorf("Error on parsing SSH private key: %s", err)
+	}
+
+	return ssh.MarshalAuthorizedKey(key.PublicKey()), nil
+}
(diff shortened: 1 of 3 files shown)


GennadySpb:
Hi @nywilken ! 
Please take a look at this PR :)

nywilken:
Hi @GennadySpb this change looks good to me. I just want to figure out if we need some documentation to denote this behavior.

nywilken (approved):
@GennadySpb thanks for this enhancement. The change does make sense to have as part of the SDK. Please not that builders will not automatically upload the generated PubKey, as the majority of the builders are set to not set any SSH public keys settings when a private key is specified. 

If you have a particular builder that you plan to use this feature with you will need to open a separate PR for it on its respective repository.

GennadySpb:
Thank you @nywilken ! I will keep it in mind.

PR #102, opened 2022-03-23
Get go-getter v2.0.2, to avoid untaring OVA files

azr:
related: 

* https://github.com/hashicorp/packer/issues/11631
* https://github.com/hashicorp/go-getter/pull/354


sylviamoss (approved):
👍🏼 🎉

PR #109, opened 2022-05-10
Update exit_code.go

oschroeder8:
This is a purely cosmetic patch to insert a space between the invalid exit code and the next sentence, which otherwise runs together.


nywilken (approved):
Good catch thanks for the fix.

PR #110, opened 2022-05-24
Bump go-getter to address vulnerabilities described in HCSEC-2022-13

nywilken:
https://discuss.hashicorp.com/t/hcsec-2022-13-multiple-vulnerabilities-in-go-getter-library/39930

PR #111, opened 2022-05-25
Update go-getter client options

nywilken:
* Disable reading and writing files through symlinks
* Add timeouts for GCS and S3 getters

Related to: https://github.com/hashicorp/packer-internal-issues/issues/38 (internal only)

### DIFF
--- a/multistep/commonsteps/step_download.go
+++ b/multistep/commonsteps/step_download.go
@@ -56,7 +56,8 @@ type StepDownload struct {
 }
 
 var defaultGetterClient = getter.Client{
-	Getters: getter.Getters,
+	Getters:         getter.Getters,
+	DisableSymlinks: true,
 }
 
 func init() {
--- a/multistep/commonsteps/step_download.go
+++ b/multistep/commonsteps/step_download.go
@@ -57,19 +57,39 @@ type StepDownload struct {
 }
 
 var defaultGetterClient = getter.Client{
-	Getters:         getter.Getters,
+	// Disable writing and reading through symlinks
 	DisableSymlinks: true,
-}
-
-func init() {
-	gcsGetter := gcs.Getter{
-		Timeout: 5 * time.Minute,
-	}
-	defaultGetterClient.Getters = append(defaultGetterClient.Getters, &gcsGetter)
-	s3Getter := s3.Getter{
-		Timeout: 5 * time.Minute,
-	}
-	defaultGetterClient.Getters = append(defaultGetterClient.Getters, &s3Getter)
+	// The order of the Getters in the list may affect the result
+	// depending if the Request.Src is detected as valid by multiple getters
+	Getters: []getter.Getter{
+		&getter.GitGetter{
+			Timeout: 5 * time.Minute,
+			Detectors: []getter.Detector{
+				new(getter.GitHubDetector),
+				new(getter.GitDetector),
+				new(getter.BitBucketDetector),
+				new(getter.GitLabDetector),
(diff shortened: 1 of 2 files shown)


sylviamoss (approved):
👍🏼 
Since this one is a bigger change that might break existing templates, do you think we should add the changelog entry in this same PR with more information about the changes? Maybe mentioning the bump as well https://github.com/hashicorp/packer-plugin-sdk/pull/110

multistep/commonsteps/step_download.go
```diff
@@ -56,12 +57,39 @@ type StepDownload struct {
...
-}
-
-func init() {
-	defaultGetterClient.Getters = append(defaultGetterClient.Getters, new(gcs.Getter))
-	defaultGetterClient.Getters = append(defaultGetterClient.Getters, new(s3.Getter))
+	// Disable writing and reading through symlinks
+	DisableSymlinks: true,
+	// The order of the Getters in the list may affect the result
+	// depending if the Request.Src is detected as valid by multiple getters
+	Getters: []getter.Getter{
```
sylviamoss, line 64: Awesome!

multistep/commonsteps/step_download.go
```diff
@@ -56,12 +57,39 @@ type StepDownload struct {
...
+		&getter.HgGetter{
+			Timeout: 5 * time.Minute,
+		},
+		new(getter.SmbClientGetter),
+		new(getter.SmbMountGetter),
+		&getter.HttpGetter{
+			Netrc:                 true,
+			XTerraformGetDisabled: true,
+			HeadFirstTimeout:      10 * time.Second,
+			ReadTimeout:           30 * time.Second,
```
picatz, line 83: The other getter timeouts are 5 minutes, while this one is essentially 30 seconds. Maybe it'd make more sense to have a uniform timeout across the getters, or somehow make these configurable?

multistep/commonsteps/step_download.go
```diff
@@ -56,12 +57,39 @@ type StepDownload struct {
...
+				new(getter.GitHubDetector),
+				new(getter.GitDetector),
+				new(getter.BitBucketDetector),
+				new(getter.GitLabDetector),
+			},
+		},
+		&getter.HgGetter{
+			Timeout: 5 * time.Minute,
+		},
+		new(getter.SmbClientGetter),
```
picatz, line 77: 🤔 Should this getter have a timeout? I see this getter only exists in v2, and doesn't have that option currently. Just curious to your thoughts.

nywilken: Yeah - I think it should. It was not caught in the go-getter updates. I will follow up with a separate PR for that change.

multistep/commonsteps/step_download.go
nywilken: I like the idea of the uniform timeout. TBH 30 seconds seems really short given the use of go-getter in Packer. I struggled a bit to find good defaults. But ultimately think that going high 30 minutes is a good base line. We can make these configurable for the step so that plugins could override if needed. 

That said there is an open issue for improving iso_config to support some addition go-getter Headers, which could also include setting some of these values. Taking a  page out of Nomad's book here.

picatz (approved):

multistep/commonsteps/step_download.go
nywilken: PR opened for the SmbClientGetter https://github.com/hashicorp/go-getter/pull/369

PR #114, opened 2022-07-02
Change step_create_cdrom_tests to pass on Windows

tempora-mutantur:
A little change that was found during setting up the project on Windows.

The tests were relying on unix paths "/" while on Windows we have "\"
and tests were failing in checkFiles with the such messages as:

unexpected file: subfolder\meta-data


nywilken (approved):
Good catch, thanks for the fix.

ChiefGhostZero0x1 (commented):
See my root finally

PR #115, opened 2022-07-06
Fix possibility to use forward slashes with nested folders in cd_files for Windows

tempora-mutantur:
Packer recommends that we use "/" as the path separator. There is a
problem with using it on Windows with nested folders because of the following behaviour:

We might pass paths with forward slashes in `StepCreateCD` struct

The string `rootFolder, err := tmp.Dir("packer_to_cdrom")` makes rootFolder use system-based slashes, which is "\\" for Windows.

This rootFolder is passed to `s.AddFile(rootFolder, toAdd)`

And a mix of two slashes ruins our replace logic while going through the `visit` function. We end up having different slashes for `allDirs` and `discardPath` in `intermediaryDirs := strings.Replace(allDirs, discardPath, "", 1)`

We change our replace login to don't depend on slashes.

The tests we had didn't cover such cases as we passed all files with paths constructed with filepath.Join(dir, fname) which returns paths with "\\" on Windows. So we added a test case to specifically check forward slashes.

test

Thank you.

### DIFF
--- a/multistep/commonsteps/step_create_cdrom.go
+++ b/multistep/commonsteps/step_create_cdrom.go
@@ -260,7 +260,9 @@ func (s *StepCreateCD) AddFile(dst, src string) error {
 		// Clean up pathing so that we preserve the base directory provided by
 		// the user but not the local pathing to that directory.
 		allDirs, base := filepath.Split(pathname)
-		intermediaryDirs := strings.Replace(allDirs, discardPath, "", 1)
+		allDirsSlashSafe := filepath.ToSlash(allDirs)
+		discardPathSlashSafe := filepath.ToSlash(discardPath)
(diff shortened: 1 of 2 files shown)


tempora-mutantur:
@nywilken Hi, very sorry for disturbing. Maybe you'll have a chance to review this

nywilken (approved):
Hi @tempora-mutantur thanks the ping here. I've been meaning to come back to this change. At first I was a little confused on why the testing was not failing locally. But I did eventually see that I need to run it with the PACKER_ACC environment 🤦 . 

This looks good to me and works as expected. I will get this in and released shortly.

tempora-mutantur:
@nywilken thank you very much

nywilken:
You got! We work on getting these changes in tested and merged sooner for the future. Thanks for making the fix. Greatly appreciated.

PR #116, opened 2022-07-11
ssh: only rely on timeout/retries if one is set

### LINKED ISSUE #87: ssh_timeout not working windows and wsl
Hello,

seems that "ssh_timeout" property is not working using packer on wsl2 and on windows.
I had set: ssh_timeout = "30m"

I had tried to find workaroud eg. setting:
ssh_proxy_host = "127.0.0.1"
then it waits 30min but it is not connecting.
I have ssh server running locally and I had added my public key to authorized keys. I'm able to ssh 127.0.01 without password.
Is there anything more that should i configure?

lbajolet-hashicorp:
In a configuration, when either the SSH timeout or the SSH handshake
retries are set, the other attribute would also be set to its default
value.

This in turn could cause a problem where a user expects his choice to be
the one that makes the final decision to abort, but in the end this
choice could cause the build to be cancelled independently as the other
setting may cause a premature cancellation.

To avoid this, we clarify the documentation, and ensure that when one is
set, the other will be ignored.

Closes: #87

### DIFF
--- a/communicator/step_connect_ssh.go
+++ b/communicator/step_connect_ssh.go
@@ -50,7 +50,10 @@ func (s *StepConnectSSH) Run(ctx context.Context, state multistep.StateBag) mult
 	}()
 
 	log.Printf("[INFO] Waiting for SSH, up to timeout: %s", s.Config.SSHTimeout)
-	timeout := time.After(s.Config.SSHTimeout)
+	timeout := make(<-chan time.Time)
+	if s.Config.SSHTimeout > 0 {
+		timeout = time.After(s.Config.SSHTimeout)
+	}
 	for {
 		// Wait for either SSH to become available, a timeout to occur,
 		// or an interrupt to come through.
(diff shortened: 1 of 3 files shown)


communicator/step_connect_ssh.go
```diff
@@ -50,7 +50,14 @@ func (s *StepConnectSSH) Run(ctx context.Context, state multistep.StateBag) mult
...
 	log.Printf("[INFO] Waiting for SSH, up to timeout: %s", s.Config.SSHTimeout)
-	timeout := time.After(s.Config.SSHTimeout)
+	var timeout <-chan time.Time
+	// Only setup a timeout when it is specified in the configs
+	if s.Config.SSHTimeout != 0 {
+		timeout = time.After(s.Config.SSHTimeout)
+	} else {
+		timeout = make(<-chan time.Time)
+	}
+
```
nywilken, line 60: Nit: we can assign the default case and drop the else
```suggestion
	timeout := make(<-chan time.Time)
	if s.Config.SSHTimeout > 0 {
		timeout = time.After(s.Config.SSHTimeout)
	} 
```

nywilken (approved):
This looks good to me. One nit.

lbajolet-hashicorp:
Good call on the nit! I'll merge this once tests pass

PR #120, opened 2022-08-07
Make ed25519 keys generated by Packer serialize in the format usable by ssh

nikolaymatrosov:
As said in the method comment, the method produced the key in a format incompatible with ssh. As a result, the key Packer SDK created resulted in the error `Load key "id_ed25519": invalid format`.
I found the [format description ](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key) and fixed the generation algorithm to fix this problem:

- I added proper padding generation
- and put the public key in the right place as described in the format.

As `ed25519` keys are getting more popular, I think it would be helpful to have them serialize correctly.

### DIFF
--- a/communicator/sshkey/generate.go
+++ b/communicator/sshkey/generate.go
@@ -62,12 +62,14 @@ func NewPair(public, private interface{}) (*Pair, error) {
 	}, nil
 }
 
-// PairFromED25519 marshalls a valid pair of openssh pem for ED25519 keypairs.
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (approved):
Hey @nikolaymatrosov,

Great job on the research (and providing the links to the docs/specs, that's very helpful in understanding the change) and the implementation. The code looks good to me, I think we can merge this right away.

Thank you very much for this contribution!

PR #123, opened 2022-08-12
Logging ssh bastion

lbajolet-hashicorp:
In order to make it easier to follow when connecting with a bastion to a SSH host, we add another layer of logging related to these actions.

Related to: https://github.com/hashicorp/packer/issues/11719

### DIFF
--- a/sdk-internals/communicator/ssh/connect.go
+++ b/sdk-internals/communicator/ssh/connect.go
@@ -2,6 +2,7 @@ package ssh
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"time"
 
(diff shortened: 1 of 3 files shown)


sdk-internals/communicator/ssh/connect.go
```diff
@@ -62,6 +63,8 @@ func BastionConnectFunc(
 			return nil, fmt.Errorf("Error connecting to bastion: %s", err)
 		}
 
+		log.Printf("[DEBUG] connected to bastion, hopping to destination")
```
nywilken, line 66: nit: hopping might not be clear for our non-native English speakers. 
```suggestion
		log.Println("[DEBUG] connected to bastion host")
		log.Println("[DEBUG] attempting connection to destination host")
```

nywilken (approved):
Nice addition. I have one nit but otherwise good to go.

sdk-internals/communicator/ssh/connect.go
lbajolet-hashicorp: Rerolled with this change, thanks!

PR #139, opened 2022-10-28
Add boot steps configuration option

jacob-carlborg:
Adds a new configuration option called `boot_steps`. This is an array of tuples of boot commands, to type when the virtual machine is booted. The first element of the tuple is the actual boot command. The second element of the tuple, which is optional, is a description of what the boot command does. This is intended to be used for interactive installers that requires many commands to complete the installation.

Both the command and the description will be printed when logging is enabled. When debug mode is enabled Packer will pause after typing each boot command. This will make it easier to follow along the installation process and make sure the Packer and the installer are in sync.

This is the SDK part of https://github.com/hashicorp/packer-plugin-qemu/pull/103.

### DIFF
--- a/bootcommand/config.go
+++ b/bootcommand/config.go
@@ -165,6 +165,45 @@ type BootConfig struct {
 	// well, and are covered in the section below on the boot command. If this
 	// is not specified, it is assumed the installer will start itself.
 	BootCommand []string `mapstructure:"boot_command"`
+	// This is an array of tuples of boot commands, to type when the virtual
+	// machine is booted. The first element of the tuple is the actual boot
+	// command. The second element of the tuple, which is optional, is a
+	// description of what the boot command does. This is intended to be used for
+	// interactive installers that requires many commands to complete the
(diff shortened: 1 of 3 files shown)


jacob-carlborg:
@lbajolet-hashicorp ping

jacob-carlborg:
@lbajolet-hashicorp No worries. I'll update the QEMU PR.

nywilken:
@lbajolet-hashicorp thanks for the ping, and @jacob-carlborg thank you for pushing this forward on the QEMU builder. I agree that this would be a nice addition to the SDK for supporting descriptive boot commands. Alternatively we can think about the usage of template_files to see how we can promote a workflow where boot commands can external to a template for reusability. 

This is not something we can prioritize at the moment so I will close this PR. But I reopened the original issue as https://github.com/hashicorp/packer-plugin-sdk/issues/216 in the SDK and marked at as a proposal for us to revisit when we are done with the bundled plugins and SDK 0.6.0 releases.

PR #147, opened 2023-01-10
ssh: fix DownloadDir empty response check

lbajolet-hashicorp:
We used to check if a ssh server returns an empty payload with a len(fi) < 0, which can never succeed as len never returns negative values.

This commit changes the condition to an == instead.


nywilken (approved):
Nice catch. Looks good to me.

PR #153, opened 2023-01-27
Remove ioutil in favour of io and os packages and exclude crypto/dsa …

rossmaclean:
Remove the ioutil package as suggested by the linter in favour of the io and os packages. Also fixes other linting warnings.

### DIFF
--- a/communicator/ssh/key_pair.go
+++ b/communicator/ssh/key_pair.go
@@ -60,7 +60,7 @@ type KeyPair struct {
 // KeyPairFromPrivateKey returns a KeyPair loaded from an existing private key.
 //
 // Supported key pair types include:
-//   - DSA
+//   - DSA (NOTE: deprecated from Go as DSA is not regarded as secure anymore, please consider RSA or ED25519 instead)
 //   - ECDSA
 //   - ED25519
 //   - RSA
--- a/communicator/ssh/key_pair.go
+++ b/communicator/ssh/key_pair.go
@@ -6,7 +6,7 @@ package ssh
 import (
 	"bytes"
 	"crypto"
-	"crypto/dsa"
+	"crypto/dsa" //nolint:all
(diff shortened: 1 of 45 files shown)


communicator/ssh/key_pair.go
```diff
@@ -60,10 +60,10 @@ type KeyPair struct {
 // KeyPairFromPrivateKey returns a KeyPair loaded from an existing private key.
 //
 // Supported key pair types include:
-// 	- DSA
-// 	- ECDSA
-// 	- ED25519
-// 	- RSA
+//   - DSA
```
lbajolet-hashicorp, line 63: I understand that DSA is marked as `nolint` here to remove the deprecation warning from the linters, which is acceptable in my opinion, however I would maybe argue in favour of warning users not to use DSA anymore since it's widely recognized as non-secure, and future versions of Go may remove the `crypto/dsa` package altogether.

Maybe consider adding the following message to this comment?

```suggestion
//   - DSA (NOTE: deprecated from Go as DSA is not regarded as secure anymore, please consider RSA or ED25519 instead)
```

lbajolet-hashicorp (approved):
Hi @rossmaclean,

Thanks for the lint pass on the codebase here, much appreciated!

I just left a comment on the DSA issue being deprecated, it wasn't clear at first why the `nolint:all` directives were added to the package imports, so I suggested leaving a warning in the documentation to make it clear why it is marked as such.

Other than that, it all looks good to me, thanks again for the PR!

nywilken (approved):
Good call on the DSA docs @lbajolet-hashicorp. LGTM

communicator/ssh/key_pair.go
rossmaclean: Good idea, makes sense to me.

rossmaclean:
Thanks both for the review.

PR #155, opened 2023-01-27
shell-local: fix supported_os error check

lbajolet-hashicorp:
Commit 34ed5d9031 which consists of linting fixes essentially had an error in the code and the error check for `supported_os' was inverted in this commit, making shell-local fail on all supported OSes now.

We re-invert this condition so it behaves as it used to.


nywilken (approved):
Looks good. Thanks for the quick fix.

PR #158, opened 2023-02-01
ssh: add warnings on DSA usage

lbajolet-hashicorp:
The DSA signature algorithm is not considered secure anymore, and is actively deprecated in the Go crypto libs (SEE https://pkg.go.dev/crypto/dsa).

To let users know that they should not use that anymore, we add a notice in the comments for the SSH private key options.

This is linked to PR #153, where the imports for the DSA package were annotated as `nolint` as importing them triggers lint warnings.


nywilken (approved):
LGTM

PR #164, opened 2023-02-24
packer-sdc: return errors on duplicate tag/field

lbajolet-hashicorp:
When generating the flattened structures for a HCL2-compatible config, we didn't prevent users from defining duplicate fields or tags, instead warning them.

The warning in itself did not prevent the resulting structures from being generated, leading into a situation where the first definition of the arg/tag would have precedence over the subsequent definitions, leading to shadowing their definitions.

To prevent this in the future, we immediately return an error when such a conflict is introduced, and signal to the user which attribute is problematic.

Related to: https://github.com/hashicorp/packer-plugin-alicloud/pull/107
Related to: https://github.com/hashicorp/packer-plugin-alicloud/pull/105


nywilken (approved):
Nice catch. This looks good. I rebased onto the latest main. I will merge once all goes green.

nywilken:
The failing windows test does not having anything to do with this change and is already fixed in a subsequent PR; merging.

PR #172, opened 2023-04-12
GHA migration finalisation

lbajolet-hashicorp:
As part of the ongoing efforts to migrate from Circle CI to GHA, this PR pins the external actions to a SHA, adds actionlint to the repo, and fixes permissions for actions that didn't have it yet.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,13 @@
+name: Lint GitHub Actions Workflows
+on:
+  push:
+  pull_request:
+permissions:
+  contents: read
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3 # v3.5.0
+      - name: "Check workflow files"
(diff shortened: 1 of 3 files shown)


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,13 @@
+name: Lint GitHub Actions Workflows
+on:
+  push:
+  pull_request:
+permissions:
+  contents: read
```
nywilken, line 6: Is contents enough for the actionlint action? I only ask because of the packages permission that was needed for the signore container.

nywilken (approved):
This looks good to me. I left a question about the permissions for the actionlint but seeing as this was tested elsewhere I just might be overly cautious because of the ghrc.io permissions.

.github/workflows/actionlint.yml
lbajolet-hashicorp: I would presume this is enough since the action only runs a command and doesn't modify the contents of the repo

lbajolet-hashicorp: While fixing the go-test.yml action, I did trigger this one, and I can confirm that with this permission alone, it works!

PR #174, opened 2023-04-19
communicator: don't set bastion cert if key is set

### LINKED ISSUE #173: ssh_certificate_file is wrongly used for ssh bastion
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### Overview of the Issue

I'm not sure if this is an issue with packer or with the proxmox plugin, but it seems that there is a bug when `ssh_certificate_file` is specified for the build image at the same time as using a bastion host with no `ssh_bastion_certificate_file`.

In this case the ssh bastion key is wrongly signed with `ssh_certificate_file` :
```
* ssh_bastion_private_key_file is invalid: failed to create cert signer: ssh: signer and cert have different public key
```

I tried explicitly setting `ssh_bastion_certificate_file` to `null` but I get the same behaviour.

The only workaround I found was to configure an ssh CA cert on the bastion host and create a cert for the bastion key with that CA.

### Packer version

- packer v1.8.6
- github.com/hashicorp/proxmox 1.1.2

### Simplified Packer Template
```
source "proxmox-clone" "vm" {
  clone_vm = "my_image"
  ...
  ssh_username = "root"
  ssh_private_key_file = "./deploy.key"
  ssh_certificate_file = "./deploy.key-cert.pub"
  ssh_bastion_username = "packer"
  ssh_bastion_host = "bastion"
  ssh_bastion_private_key_file = "bastion.key"
  ssh_bastion_certificate_file = null
}
```

### Log Fragments and crash.log files

```
* ssh_bastion_private_key_file is invalid: failed to create cert signer: ssh: signer and cert have different public key
```

lbajolet-hashicorp:
When attempting to set the bastion key/certificate for authenticating with the bastion, we generally fallback to the ones defined by the SSH configuration.

However, if the bastion SSH key is set, and not the certificate, but the SSH connection's are, since the conditions are separate, we end-up in a situation where the bastion's SSH key uses the one from the config, and the certificate fall backs to the one from the SSH connection.

This in turn fails, as the certificate's public key matches the private key from the SSH connection, and not the bastion's.

To avoid a situation like this, we only fallback to the SSH connection's certificate if the bastion's SSH key isn't set.

Closes #173

### DIFF
--- a/communicator/config.go
+++ b/communicator/config.go
@@ -496,6 +496,16 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 		c.SSHKeepAliveInterval = 5 * time.Second
 	}
 
+	// Validation
+	var errs []error
+	if c.SSHPrivateKeyFile == "" && c.SSHCertificateFile != "" {
+		errs = append(errs, fmt.Errorf("ssh_private_key_file must be specified if ssh_certificate_file is specified"))
+	}
+
+	if c.SSHBastionPrivateKeyFile == "" && c.SSHBastionCertificateFile != "" {
+		errs = append(errs, fmt.Errorf("ssh_bastion_private_key_file must be specified if ssh_bastion_certificate_file is specified"))
+	}
+
 	if c.SSHBastionHost != "" {
 		if c.SSHBastionPort == 0 {
 			c.SSHBastionPort = 22
@@ -503,12 +513,8 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 
 		if c.SSHBastionPrivateKeyFile == "" && c.SSHPrivateKeyFile != "" {
 			c.SSHBastionPrivateKeyFile = c.SSHPrivateKeyFile
-		}
-
-		if c.SSHBastionCertificateFile == "" && c.SSHCertificateFile != "" {
 			c.SSHBastionCertificateFile = c.SSHCertificateFile
 		}
-
 	}
 
 	if c.SSHProxyHost != "" {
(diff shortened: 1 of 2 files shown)


lbajolet-hashicorp:
NOTE: I'm still working on tests right now, so this may not be ready to be merged immediately, but this can be reviewed in the meantime

communicator/config.go
```diff
@@ -503,12 +503,12 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 
 		if c.SSHBastionPrivateKeyFile == "" && c.SSHPrivateKeyFile != "" {
 			c.SSHBastionPrivateKeyFile = c.SSHPrivateKeyFile
-		}
 
-		if c.SSHBastionCertificateFile == "" && c.SSHCertificateFile != "" {
-			c.SSHBastionCertificateFile = c.SSHCertificateFile
+			// Only try to set the bastion certificate to the ssh certificate if the bastion private key isn't set
```
nywilken, line 507: I can see a mismatch here if a user provides a BastionCertFile but no BastionKey. In this case the SSHPrivateKeyFile will be used as SSHBastionPrivateKeyFile but the cert will be unmodified. There looks to be a validation check to ensure a bastion key file is set but if that changes it could be an issue. 

```go
		if (c.SSHBastionPrivateKeyFile == "" && c.SSHBastionCertificateFile == "") && c.SSHPrivateKeyFile != "" {
			c.SSHBastionPrivateKeyFile = c.SSHPrivateKeyFile
			c.SSHBastionCertificateFile = c.SSHCertificateFile
		}
```

lbajolet-hashicorp: Hm, in the case where someone provides a cert and no key, this should always be an error rather than trying to fallback.
I'll add this case to the test suite and make sure it errors. This should also be the case for the host SSH connection, and there doesn't seem to be any check for this over here, if the host SSH key isn't set, the host cert is ignored.

nywilken (approved):
LGTM

PR #181, opened 2023-04-28
ssh: improve debug logging

lbajolet-hashicorp:
When trying to figure out the exact behaviour of the ssh communicator when copying directories, I found myself wondering what each log was conveying.

To make it clearer what's happening on each step, we add some more logs during the copy steps, and make the `creating directory' log a bit more explicit, so we understand what's happening.

Related to: #180


nywilken (approved):
LGTM

PR #184, opened 2023-05-25
Fix menu boot command vnc key stroke

korli:
`<menu>` would output individual keys instead of the menu special key.


lbajolet-hashicorp (approved):
LGTM, good catch on that one @korli, thanks for the fix!

korli:
> LGTM, good catch on that one @korli, thanks for the fix!

Good! Let me know if commits need to be squashed.

lbajolet-hashicorp:
No need to worry about that, Github can do it for us, I'm merging this now!

Issue #135, opened 2022-10-18
Upgrade github.com/zclconf/go-cty to v1.11.1


nywilken:
Upgrade github.com/zclconf/go-cty to v1.11.1

In v1.11.0 go-cty dropped all `encoding/gob` support - refer to [commit](https://github.com/zclconf/go-cty/commit/97bafac0dea33a3f74db88ab54dea2937b9e8aef) for details. 
Gob support is currently being used by the Datasource RPC server, and possiblity within other locations, for communication with the data source plugin components. Before upgrading
to the latest go-cty version the RPC serialization code must be updated to use a new serialization format or we can look into copying the encoding/gob support into 
the Packer SDK if there are no other options.

NorseGaud:
bump

lbajolet-hashicorp:
Hi @NorseGaud, sorry for taking a long time to address this.

We're still evaluating the kind of change needed for this update to be applicable to Packer, as Wilken mentions in the description for this issue, this change of the cty library removes support for `gob` encoding, which is what we use for the Packer->plugin communication, so this'll require some refactoring before we can bump it.

If possible, we want to avoid breaking communication between Packer and older versions of plugins that are installed, so we need to tread carefully, this will likely take some time before we land that change in the SDK, Packer, and the plugins.

We'll keep this issue updated with more information as we proceed with this change, and let you know when it's safe to bump cty everywhere.

Sorry for not bringing better news on that end, hopefully this won't be a blocker for you.

nywilken:
In v1.11.0 of [github.com/zclconf/go-cty](https://github.com/zclconf/go-cty/blob/main/CHANGELOG.md#upgrade-notes-1), encoding/gob support was removed from the package. The Packer plugin SDK relies on gob for supporting HCL2 templates. The removal of the encoding causes Packer to crash when using HCL2 templates.


The team and I are working internally to find the best path forward, as a change to the wire protocol from gob to something else will introduce a breaking change and plugin incompatibilities. We are working to notify plugin developers of the incompatible go-cty package and ask that all external plugins pin to the versions of the following two packages to those used by the Packer plugin SDK.

[github.com/hashicorp/hcl/v2](http://github.com/hashicorp/hcl/v2) v2.13.0
[github.com/zclconf/go-cty](http://github.com/zclconf/go-cty) v1.10.0


 Feel free to reach out if you have any questions or concerns.

nywilken:
[...]

PR #186, opened 2023-06-13
[HPR-1108] Add compile time check for go-cty

nywilken:
In v1.11.0, the go-cty package dropped support for encoding/gob. Gob support is used
by the SDK to serialize HCL2 object specs over the wire. If a plugin or Packer upgrades
their version of zclconf/go-cty to one that does not contain Gob support the plugin will build
but crash when trying to run a build on an HCL2 template. By adding this check, a consumer of the SDK
will fail to compile if they are using an unsupported version of go-cty. This check is being added
as a guard to prevent Packer SDK consumers from becoming out of sync with unsupported go-cty versions.

```
~> go get github.com/hashicorp/go-cty v1.13.0
~> make dev
rpc/cty_encode.go:15:24 cannot use cty.Value{} (value of type cty.Value) as type gob.GobEncoder
```

- Related to https://github.com/hashicorp/packer-plugin-sdk/issues/135
- Related to https://github.com/hashicorp/packer/issues/12182

### TODO

- [x] Open Issue for Compile Time Check Workaround - https://github.com/hashicorp/packer-plugin-sdk/issues/187

### DIFF
--- a/rpc/cty_encode.go
+++ b/rpc/cty_encode.go
@@ -4,9 +4,15 @@
 package rpc
 
 import (
+	"encoding/gob"
+
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/json"
 )
+
+// Test that cty types implement the gob.GobEncoder interface.
+// Support for encoding/gob was removed in github.com/zclconf/go-cty@v1.11.0.
+var _ gob.GobEncoder = cty.Value{}
 
 // cty.Value is does not know how to encode itself through the wire so we
 // transform it to bytes.
--- a/rpc/cty_encode.go
+++ b/rpc/cty_encode.go
@@ -4,15 +4,9 @@
 package rpc
 
 import (
-	"encoding/gob"
-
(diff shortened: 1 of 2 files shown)


JenGoldstrich (approved):
LGTM!

sylviamoss (approved):
I am still amazed by this simple way to prevent plugins to upgrade to an incompatible gocty version. 🎉

rpc/cty_encode.go
```diff
@@ -4,10 +4,16 @@
...
 import (
+	"encoding/gob"
+
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/json"
 )
 
+// Test that cty types implement the gob.GobEncoder interface.
+// Support for encoding/gob was removed in github.com/zclconf/go-cty@v1.11.0.
+var _ gob.GobEncoder = cty.Value{}
```
sylviamoss, line 15: Is it possible to put this inside an `init` method so you can check the type and pop an error message of our choice to include the related issue link? 🤔

Issue #187, opened 2023-06-14
cty.Value does not implement gob.GobEncoder


nywilken:
## Related Errors
- cannot use cty.Value{} (value of type cty.Value) as gob.GobEncoder value in variable declaration
- Datasource.OutputSpec failed: gob: type cty.Type has no exported fields

## What's this error?

```
~> go get github.com/hashicorp/go-cty v1.13.0
~> make dev
packer-plugin-sdk/rpc/init.go:20:24: cannot use cty.Value{} (value of type cty.Value) 
as gob.GobEncoder value in variable declaration: cty.Value does not implement gob.GobEncoder 
(missing method GobEncode)
```

In v1.11.0, the github.com/zclconf/go-cty package dropped support for `encoding/gob`. Gob support is used
by the Packer SDK to serialize HCL2 object specs over the wire. If a plugin or Packer upgrades
their version of github.com/zclconf/go-cty to one that does not contain Gob support (v1.11.0+) the plugin will build
but crash when trying to run a build on an HCL2 template. 

The error being reported above is due to a compile type check added to the Packer SDK to alert a consumer of the SDK
that they are using an unsupported version of github.com/zclconf/go-cty. The `gob.GobEncodercheck` was added
as a guard to prevent Packer SDK consumers from becoming out of sync with unsupported go-cty versions.

## How do I fix it?
If you are seeing this error then you are running a release of the Packer plugin SDK that now enforces a type check
for the `encoding/gob` support in github.com/zclconf/go-cty. There are a number of ways to fix this issue, which are all meant to be temporary fixes until v1.0.0 of the Packer Plugin SDK, which will move from using `encoding/gob` for serializing cty Values. 

In an effort to provide Packer plugins access to the github.com/zclconf/go-cty enhancements, while still providing access go `encoding/gob` support the Packer team has forked the go-cty package under github.com/nywilken/go-cty. This is a temporary fork that will be removed once it is no longer needed. Until then plugin developers are encourage to use the fork as a replacement (within the go.mod file) for github.com/zclconf/go-cty. 

For a complete list of changes within the go-cty fork for Packer refer to the [CHANGELOG](https://github.com/nywilken/go-cty/blob/main/CHANGELOG.md)

If you are using a version of the Packer Plugin SDK prior to the release of the compile time check within the SDK
you can simply use the recommended replace directive (Option 1) to pull in the fork and gain access to the latest enhancements to go-cty.

#### Option 1: Manual Replace Directive
[...]

PR #189, opened 2023-06-20
[HPR-1108] Replace gitub.com/zclconf/go-cty@1.12.1 with fork github.com/nywilken/go-cty

nywilken:
In an effort to provide Packer plugins access to the github.com/zclconf/go-cty enhancements, while still providing access to encoding/gob support the Packer team has forked the go-cty package under github.com/nywilken/go-cty. 
This is a temporary fork that will be removed once it is no longer needed.

* This change also reverts the Go statement to 1.18 in the go.mod file. The SDK has no dependencies on Go 1.20 to justify the Go version bump in the module file.

Relates to: https://github.com/hashicorp/packer-plugin-sdk/issues/187

### DIFF
--- a/filelock/filelock.go
+++ b/filelock/filelock.go
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
 //go:build !solaris
-// +build !solaris
 
 package filelock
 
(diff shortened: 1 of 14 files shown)


lbajolet-hashicorp (approved):
Easy enough to review, I like that :)
LGTM!

PR #190, opened 2023-06-22
[HPR-1190] Add `packer-sdc fix` command

### LINKED ISSUE #135: Upgrade github.com/zclconf/go-cty to v1.11.1

nywilken:
In an effort to automate the temporary fix, and its removal in the future, a fix sub-command has been added to the `packer-sdc` command; not to be confused with `packer fix` which fixes legacy JSON template files. 

By running the `packer-sdc fix` command within the plugin's root project  directory the command will check the plugin's go.mod file has a replace directive to the go-cty fork.  If no replace directive is found it will be automatically added to the go.mod file.

Depends on #189
Relates to: #187 
Closes #135 

#### Help Output
```shell
~>  ./packer-sdc fix -h

Usage: packer-sdc fix [options] directory

  Fix rewrites parts of the plugin codebase to address known issues or
  common workarounds used within plugins consuming the Packer plugin SDK.

Options:
  -diff         If the -diff flag is set fix prints the differences an applied fix would introduce.

Available fixes:
  gocty         Adds a replace directive for github.com/zclconf/go-cty to github.com/nywilken/go-cty
```

#### Basic Diff
```shell
~>  ./packer-sdc fix -diff ../../../packer/
--- /home/vagrant/Development/packer/go.mod
+++ /home/vagrant/Development/packer/go.modfixed
@@ -213,3 +213,5 @@
 )

 go 1.20
+
+replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v1.12.1 // added by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187
```

#### Applied Fix
```
~>  ./packer-sdc fix ../../../packer/
/home/vagrant/Development/packer/go.mod
~>  git diff
diff --git a/go.mod b/go.mod
index fa48d7d58..02b455992 100644
--- a/go.mod
+++ b/go.mod
@@ -213,3 +213,5 @@ require (
 )

 go 1.20
+
+replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v1.12.1 // added by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187

### CODE REVIEW

File: cmd/packer-sdc/internal/fix/fix.go [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+
+func (cmd *Command) Flags() *flag.FlagSet {
+	fs := flag.NewFlagSet(cmdPrefix, flag.ExitOnError)
+	fs.BoolVar(&cmd.Check, "check", false, "Check plugin for potential fixes [dry-run].")
+	return fs
+}
+
+// Help displays usage for the command.
+func (cmd *Command) Help() string {
+	return "\n" + readme
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Alternatively I'd think we can remove the first line before printing the message through some slicing magic or a well-placed regexp replace:

```go
var rmHeader = regexp.MustCompile("##\\s*`fix`\\s*")

// Help displays usage for the command.
func (cmd *Command) Help() string {
    return rmHeader.ReplaceAllString(readme, "")
}
```

Or:

```go
func (cmd *Command) Help() string {
    return strings.split("\n", readme)[1:]
}
```

cmd/packer-sdc/internal/fix/gocty.go
```diff
@@ -0,0 +1,124 @@
...
+		if r.New.Path != f.NewPath {
+			return false, errors.New("found unexpected replace for " + r.Old.Path)
+		}
+		return r.New.Version != f.modFileFormmatedVersion(), nil
+	}
+	return true, nil
+}
+
+func (f GoCtyFix) Fix(dir string, check bool) error {
+	ok, err := f.isUnfixed(dir)
```
lbajolet-hashicorp, line 88: While I understand the intent, I'm not sure `ok` is the best name for this, maybe consider naming it `unfixed` to make it clear what it is we're testing on?

cmd/packer-sdc/internal/fix/gocty.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return nil
+	}
+	if check {
+		fmt.Printf("%s %5s\n", "gocty", "Unfixed!")
```
lbajolet-hashicorp, line 96: What's this `%5s`? Also since everything is static here, why the format-string? We could as well use `fmt.Println` here I'd think.

```suggestion
		fmt.Println("gocty Unfixed!")
```

Also regarding the phrasing, I think we'll confuse users by printing `unfixed`, my understanding of the `check` is to decide if we need to fix the `go.mod`, so I would suggest amending the message to clearly state that they need to run it in fix mode, something like `[packer-sdc] gocty needs to be fixed`.

cmd/packer-sdc/internal/fix/gocty.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	mf, err := modfile.Parse(modFilePath, data, nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse plugin go.mod file")
+	}
+
+	if err := mf.DropReplace(f.OldPath, ""); err != nil {
+		return errors.Wrap(err, "failed to apply gocty fix")
+	}
+	commentSuffix := " // add by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187"
```
lbajolet-hashicorp, line 117: Nit

```suggestion
	commentSuffix := " // added by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187"
```

lbajolet-hashicorp (commented):
Couple nits, but the code looks good to me overall!

cmd/packer-sdc/internal/fix/gocty.go
nywilken: Good call out. I made the suggested change.

cmd/packer-sdc/internal/fix/gocty.go
nywilken: I replaced this check with a basic diff output for now. It currently looks like the following
```shell
~>  ./packer-sdc fix -diff ../../../packer/
/home/vagrant/Development/packer/go.mod:
  []uint8(
        """
        ... // 213 identical lines

        go 1.20
+
+       replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v1.12.1 // add by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187
+
        """
  )
```

cmd/packer-sdc/internal/fix/fix.go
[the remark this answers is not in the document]
nywilken: Yeah I'm not a fan of this whole readme for the help text. I changed for the fix command and will change for all the other commands. let me know what you think. I will go with something similar to what we have for Packer.

cmd/packer-sdc/internal/fix/cmd.go
```diff
@@ -0,0 +1,153 @@
...
+func (cmd *Command) run(args []string) error {
+	f := cmd.Flags()
+	err := f.Parse(args)
+	if err != nil {
+		return errors.Wrap(err, "unable to parse flags for fix command")
+	}
+
+	if f.NArg() != 1 {
+		fmt.Println(cmd.Help())
+		return nil
```
nywilken, line 73: This should be an error. I will rework the error messaging.

cmd/packer-sdc/internal/fix/cmd.go
```diff
@@ -0,0 +1,153 @@
...
+	fs := flag.NewFlagSet(cmdPrefix, flag.ExitOnError)
+	fs.BoolVar(&cmd.Diff, "diff", false, "if set prints the differences a rewrite would introduce.")
+	return fs
+}
+
+// Help displays usage for the command.
+func (cmd *Command) Help() string {
+	var s strings.Builder
+	for _, fix := range availableFixes {
+		s.WriteString(fmt.Sprintf("  %s\t\t%s", fix.name, fix.description))
```
lbajolet-hashicorp, line 38: I believe you should terminate this helpstring with a `\n` so we can have several not on the same line.

```suggestion
		s.WriteString(fmt.Sprintf("  %s\t\t%s\n", fix.name, fix.description))
```

cmd/packer-sdc/internal/fix/gocty.go
```diff
@@ -0,0 +1,115 @@
...
+		NewPath:    newPath,
+		NewVersion: newVersion,
+	},
+}
+
+type GoCtyFix struct {
+	OldPath, NewPath, NewVersion string
+}
+
+func (f GoCtyFix) modFileFormmatedVersion() string {
```
lbajolet-hashicorp, line 36: Nit

```suggestion
func (f GoCtyFix) modFileFormattedVersion() string {
```

cmd/packer-sdc/internal/fix/gocty.go
```diff
@@ -0,0 +1,115 @@
...
+	return fmt.Sprintf("v%s", f.NewVersion)
+}
+
+func (f GoCtyFix) isUnfixed(modFilePath string, data []byte) (bool, error) {
```
lbajolet-hashicorp, line 40: Coming back on this function, I'm a bit confused when looking at the name since I interpret it as a fix that got removed, not a fix that needs to happen.
May I suggest calling it `needsFix`? That would make it clearer imho

cmd/packer-sdc/internal/fix/fix.go
```diff
@@ -0,0 +1,26 @@
...
+package fix
+
+// Fixer applies all defined fixes on a plugin dir; a Fixer should be idempotent.
+type fixer interface {
+	Fix(filename string, data []byte) ([]byte, error)
+}
+
+type fix struct {
+	name, description string
+	fileGlob          string
```
lbajolet-hashicorp, line 13: I wonder, do we need this one? It feels a bit limiting for future fixes maybe since globs may not recursively look for files to glob, and right now it's completely static, so I wonder if we need to define this as part of `fix` right now

lbajolet-hashicorp (commented):
Just left a couple of nits/questions, but the current implementation looks good to me.
I see this is still a draft so I'll wait until this is final to approve, but I'd think this is almost ready to merge.

cmd/packer-sdc/internal/fix/fix.go
nywilken: So for each of the fixers I'm looking to in to help identify what file(s) to apply the fix on. Do you have a suggest on how I should handle? Or are you suggesting just statically pointing to the file since it is just one file in the project at this time?

lbajolet-hashicorp: I understand the objective for this attribute, but I think that for now, this is maybe too early to introduce this glob as the way to get the file(s) to work on, and I am worried it may be limited in the end.
Maybe this could be a function instead if we want to be able to add a new fixer without having to specify something similar later? I think a function would be more futureproof for this.

nywilken: Ah I like this so maybe a function that takes a directory and returns the files to apply the fix on. Is that what you are thinking?

lbajolet-hashicorp: Essentially yes!

nywilken: Done tell me what you think?

cmd/packer-sdc/internal/fix/README.md
```diff
@@ -0,0 +1,12 @@
...
+Fix rewrites parts of the plugin codebase to address known issues or common workarounds used within plugins consuming the Packer plugin SDK.
+
+Options:
+
+ -diff    If the -diff flag is set, no files are rewritten. Instead, fix prints the differences a rewrite would introduce.
+
+Available Fixes:
+
+ gocty
+           Adds a replace directive for github.com/zclconf/go-cty to github.com/nywilken/go-cty
```
sylviamoss, line 12: Nice readme! Should the issue be linked here as well?

cmd/packer-sdc/internal/fix/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+	for _, f := range availableFixes {
+		matches, err := f.scan(rootDir)
+		if err != nil {
+			return fmt.Errorf("failed to apply %s fix: %s", f.name, err)
+		}
+
+		//matches contains all files to apply the said fix on
+		for _, filename := range matches {
+			if _, ok := srcFiles[filename]; !ok {
+				// TODO handle error
```
sylviamoss, line 117: There's a TODO here

cmd/packer-sdc/internal/fix/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+				bs, _ := os.ReadFile(filename)
+				srcFiles[filename] = bytes.Clone(bs)
+			}
+
+			fixedData, ok := fixedFiles[filename]
```
sylviamoss, line 122: Why you fix the data again?

cmd/packer-sdc/internal/fix/gocty.go
```diff
@@ -0,0 +1,114 @@
...
+		if r.New.Version == f.modFileFormattedVersion() {
+			return data, nil
+		}
+	}
+
+	if err := mf.DropReplace(f.OldPath, ""); err != nil {
+		return nil, fmt.Errorf("%s: failed to drop previously added replacement fix %v", modFilePath, err)
+	}
+
+	commentSuffix := " // added by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187"
```
sylviamoss, line 104: Nice!

cmd/packer-sdc/internal/fix/testdata/norequire/go-cty/go.mod
```diff
@@ -0,0 +1,19 @@
+module example
```
sylviamoss, line 1: I think you meant to name the filter `noreplace` instead of `norequire` right?

cmd/packer-sdc/internal/fix/fix.go
lbajolet-hashicorp: Looks good to me!

cmd/packer-sdc/internal/fix/cmd.go
lbajolet-hashicorp: Should this still be changed? I assume returning the error will take care of printing the help function, doesn't it?

lbajolet-hashicorp (approved):
@sylviamoss found some good things to point out, and there's still the outstanding error comment that might need addressing, but overall this looks good to me!

I'm approving this right now, when you addressed the remaining comments this is good to go imho.

cmd/packer-sdc/internal/fix/cmd.go
nywilken: Should this be changed? Maybe, as a user I would expect this to exit and return a non-zero exit code. I will probably move this into the Run function for the reason below.

> I assume returning the error will take care of printing the help function, doesn't it?
Yes, but the error message gets printed out using log.Printf so the usage text looks strange. As I right this I will just drop the log.Printf in favor of a fmt.Printf and return the help message as the error message.

cmd/packer-sdc/internal/fix/README.md
nywilken: Good call out. I will add that to the README.

cmd/packer-sdc/internal/fix/cmd.go
nywilken: So it is not really an issue since we have one fix but if there are multiple fixes that would get applied to the same file  we need to make sure that the fixes are accumulative and not modifying the same unfixed src file. 

To do ☝️ I store the already fixed data into the fixedFiles map and before applying another fix I check if the contents of the file have changed if so I grab the fixed file before applying the next fix. If we've never seen the file before or it has not been fixed previously we copy the original source into fixedData and only store within the fixedFiles map if a fix has been applied. 

Please let me know if I explained that well or if you're still unclear on why

cmd/packer-sdc/internal/fix/cmd.go
nywilken: This is the new error message

```shell
~>  ./packer-sdc fix
packer-sdc fix: missing directory argument

Usage: packer-sdc fix [options] directory

  Fix rewrites parts of the plugin codebase to address known issues or
  common workarounds used within plugins consuming the Packer plugin SDK.

Options:
  -diff         If the -diff flag is set fix prints the differences an applied fix would introduce.

Available fixes:
  gocty         Adds a replace directive for github.com/zclconf/go-cty to github.com/nywilken/go-cty
```

lbajolet-hashicorp: LGTM!

cmd/packer-sdc/internal/fix/cmd.go
nywilken: TODO handled. Thanks for catching that.

cmd/packer-sdc/internal/fix/cmd.go
sylviamoss: oh yeah, I see now! First time I looked I didn't realized that fixedFiles is used for all fixes. Makes sense to me, thanks for explaining!

sylviamoss (approved):
Very good command, ready to implement more fixes. 👍🏼  LGTM!

PR #191, opened 2023-06-26
Remove extra backticks from HTTP server documentation

edigaryev:
Problem: HTTP server example usage is shown with backticks [in the documentation](https://developer.hashicorp.com/packer/plugins/builders/tart#http-server-configuration) ([source includes `packer-plugin-sdk/multistep/commonsteps/HTTPConfig.mdx`](https://github.com/cirruslabs/packer-plugin-tart/blob/c5e6e6f40fafb0fc033e5048fd706721d7fe2869/docs/builders/tart.mdx#L61)):

Screenshot 2023-06-26 at 18 45 33

Solution: explicitly treat example usage as code.


edigaryev:
Ran the `make generate`, PTAL.

PR #193, opened 2023-06-30
Add Support for Block Labels when using `mapstructure-to-hcl2`

aidan-mundy:
These changes add support for Block Labels to `mapstructure-to-hcl`. Any string (or string pointer) field on a struct that also has the "label" option on the hcl struct tag (like this `'mapstructure:"nameLabel" hcl:",label"'`, except with "`" instead of "'") will be treated as a label, and the BlockLabel schema will automatically generate.

Creating a label on the root-level Config struct for a component will trigger [a crash](https://gist.github.com/aidan-mundy/0a7c1a6a653b7235ae0ad1d584fe59ba) before the build begins. This happens because labels at that level are managed by another component of Packer. I consider this to be expected behavior because a label is not valid in that location.


cmd/packer-sdc/internal/mapstructure-to-hcl2/mapstructure-to-hcl2.go
```diff
@@ -52,6 +53,8 @@ var (
 	readme string
 )
 
+const HCLLABELINDEXKEY = "hcllabelindex"
```
JenGoldstrich, line 56: I'm not sure if this suggestion is idiomatic but I had trouble reading this variable name on my first time through (I kept reading it as Belinda Key 🤦‍♀️) I'd recommend naming it HCL_LABEL_INDEX_KEY

JenGoldstrich (commented):
This is looking good to me Aidan! I think this approach makes sense to me, I think we should add testing for this in the existing units, like we talked about in slack I think it'd be good to have an example on your provisioners you're working on since that'd make manual testing of this super easy

We'll likely also want to update https://github.com/hashicorp/packer-plugin-scaffolding the scaffolding to have an example of this, but thats after this is released of course

PR #197, opened 2023-07-31
Fix replace go mod

lbajolet-hashicorp:
Release 0.5.0 of the SDK was borked as the `replace` statement for `go-cty` in the `go.mod` makes it incompatible with `go install`, which is how dependents of the SDK install `packer-sdc`.

To fix this, we remove this statement from the `go.mod`, and amend the test action so that it adds that statement before testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/hashicorp/packer-plugin-sdk
 
+require github.com/zclconf/go-cty v1.12.1 // go-cty v1.11.0 removed gob encoding support so it cannot work with the Packer SDK as-is, you need to run `packer-sdc fix .' to change that
+
 require (
 	cloud.google.com/go v0.105.0 // indirect
 	cloud.google.com/go/storage v1.27.0 // indirect
@@ -63,7 +65,6 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/ugorji/go/codec v1.2.6
 	github.com/ulikunitz/xz v0.5.10 // indirect
-	github.com/zclconf/go-cty v1.12.1
 	golang.org/x/crypto v0.0.0-20220517005047-85d78b3ac167
 	golang.org/x/mobile v0.0.0-20210901025245-1fde1d6c3ca1
 	golang.org/x/mod v0.8.0


.github/workflows/go-test.yml
```diff
@@ -25,22 +25,33 @@ jobs:
...
+      # Required for passing tests as we need to have a replace in the go.mod
+      # otherwise we'll get compile-time errors when running them
+      - name: Fix gocty
+        run: |
+          make install-gen-deps
+          packer-sdc fix .
+          go mod tidy
+          git config user.name "Packer CI"
+          git config user.email "ci@packer.io"
+          git add go.mod go.sum && git commit -m 'tmp changes for testing' # we need to add them as a commit to avoid `generate-check` reporting diffs
```
nywilken, line 49: Seeing as you are adding this step here to avoid the generation check what do you think about not doing this so early and just making it part of the test only. Preferably as part of the make test target so that locally we run packer-sdc fix as well for testing.

Does the generation change if the replace is in place?

go.mod
```diff
@@ -1,8 +1,5 @@
 module github.com/hashicorp/packer-plugin-sdk
 
-// As described in https://github.com/hashicorp/packer-plugin-sdk/issues/187
-replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v1.12.1
```
nywilken, line 4: Seeing as we are validating the replace during testing to check for potential issue what do you think about not bumping github.com/zclconf/go-cty and just keeping it pinned to 1.10.0? 

Will that cause issues?

nywilken: This def limits the version of Hcl/v2 to as hcl/v2@2.15.0 requires at least 1.12.0 of zclconf/go-cty

lbajolet-hashicorp: Issues no, but that will not make the compile-time error appear, so users won't be hinted at running `packer-sdc fix .` when they upgrade

nywilken: I'm talking about just in the SDK. The compile time check will fail if a user tries to include the SDK with an unsupported version of go-cty. So they should still see the hint.

nywilken (approved):
This looks good to me. Can we rebase the commits before merging.

PR #199, opened 2023-08-01
go.mod: retract v0.5.0

lbajolet-hashicorp:
Since the SDK in v0.5.0 contains a release statement in its go.mod, this makes it impossible to go install packer-sdc from a plugin.

Since those plugins rely on their Makefile to do this, we cannot change them all at once, so we fixed the problem in the SDK code itself, and v0.5.0 is broken, so we retract it in this commit.


nywilken (approved):
This looks good. Please rebase to fix conflicts.

PR #200, opened 2023-08-01
Bump go-getter to v2.2.1

zliang-akamai:
v2.2.0 always trigger our antivirus alert...

And it was fixed in v2.2.1 via https://github.com/hashicorp/go-getter/pull/425


nywilken (approved):
Hi @zliang-akamai thanks for this contribution. The change looks good and is warranted. I'm approving but will not merge immediately as I would like to give the 0.5.1 release the weekend to see if any issues are reported. This will be part of the next release 0.5.2 shortly after the pausing period.

zliang-akamai:
@nywilken Yep, no need to rush. Thank you for the review!

PR #202, opened 2023-08-22
acctest: add capability to add args to build

lbajolet-hashicorp:
When running an acceptance test, we may need to invoke Packer with extra flags to ensure it behaves as expected.

Therefore, we add this capability to the TestCase structure so it injects them verbatim in the command-line.

### DIFF
--- a/acctest/pluginacc.go
+++ b/acctest/pluginacc.go
@@ -25,6 +25,10 @@ const TestEnvVar = "PACKER_ACC"
 type PluginTestCase struct {
 	// Init, if true `packer init` will be executed prior to `packer build`.
 	Init bool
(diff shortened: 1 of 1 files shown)


acctest/pluginacc.go
```diff
@@ -25,6 +25,9 @@ const TestEnvVar = "PACKER_ACC"
 type PluginTestCase struct {
 	// Init, if true `packer init` will be executed prior to `packer build`.
 	Init bool
+	// BuildExtraArgs is the list of arguments to be passed as extra to the packer build
+	// command, before `--machine-readable`
```
nywilken, line 29: Good call. I do wonder if we should just state that by default `-machine-readable` is auto appended to this list to prevent someone from adding it twice.

nywilken (approved):
LGTM

PR #204, opened 2023-08-25
hcl2helper: preemptively panic on nil hcl spec

lbajolet-hashicorp:
When the code from a plugin differs from its generated FlatConfig, we may encounter a nil object spec for the values in the output of a component.
This translates as a nil dereferencing panic within the hcl library, caused by a config and tags being out-of-sync.

To avoid users the hassle of delving in this code, and the hcl library's, we preemptively panic in the SDK code, with a message suggesting regenerating the code for the component's specs.


JenGoldstrich (approved):
This makes sense to me, if spec is null at this point we might as well panic with a useful error instead of a NRE

nywilken (approved):
LGTM - Nice catch.

PR #225, opened 2024-03-12
Ui: add format-string alternatives to functions

lbajolet-hashicorp:
The say/error/ask methods on Ui only accept formatted strings, which is a bit cumbersome to do at callsite every time we need to print out something formatted with dynamic information.

To reduce that cumbersomness, we add some convenience alternatives to the Ui implementations in the SDK, so they now expose Sayf, Askf and Errorf in addition to the rest.


nywilken (approved):
Nice addition. This looks good to me.

PR #228, opened 2024-03-21
Version parsing rehaul

lbajolet-hashicorp:
As it stands now, plugin developers can only specify a version's core semver-compatible version, and a prerelease.
In usage though, some plugins use metadata to add some extra context on the binary that was built, like the git commit, or the delta between the last release and the current HEAD from which a plugin was built. 

This, coupled with the 1.11.0 changes to plugin loading and version support, means that we cannot support such a workflow with the current code, as we will now start enforcing proper semver for plugins, so we are at risk of having plugins either not load, or lose information when releasing.

Therefore as an attempt to address those issues, we are adding official support in the SDK for metadata in versions. That change introduces two new functions: `NewPluginVersion` and `NewRawVersion`. Both are intended as replacements for the now deprecated `InitializePluginVersion`.

In addition to this change, we are also limiting the number of segments to 3 in the version number, there are too many unknowns around 4-segmented version numbers, in addition to this not being semver compliant (despite the library supporting them).

cc @NorseGaud @torarnv @chris-rock since we've had recent exchanges regarding plugins recently on your respective repositories, this PR should cement the addition of metadata to what is supported for versions. Please feel free to take a look at this code, and let me know your thoughts, I'd like to know if this kind of interface would suit your development practices.
Thanks in advance!

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -47,6 +47,11 @@ func InitializePluginVersion(vers, versionPrerelease string) *PluginVersion {
 // As NewRawVersion, if the version is invalid, it will panic.
 func NewRawVersion(rawSemVer string) *PluginVersion {
 	vers := version.Must(version.NewVersion(rawSemVer))
+
+	if len(vers.Segments()) != 3 {
+		panic(fmt.Sprintf("versions should only have 3 segments, %q had %d", rawSemVer, len(vers.Segments())))
+	}
+
 	return &PluginVersion{
 		version:           vers.Core().String(),
 		versionPrerelease: vers.Prerelease(),
@@ -81,16 +86,7 @@ func NewPluginVersion(vers, versionPrerelease, versionMetadata string) *PluginVe
 		versionRawString = fmt.Sprintf("%s+%s", versionRawString, versionMetadata)
 	}
 
-	// This call initializes the SemVer to make sure that if Packer crashes due
-	// to an invalid SemVer it's at the very beginning of the Packer run.
-	semVer := version.Must(version.NewVersion(versionRawString))
-
-	return &PluginVersion{
-		version:           semVer.Core().String(),
-		versionPrerelease: semVer.Prerelease(),
-		versionMetadata:   semVer.Metadata(),
-		semVer:            semVer,
-	}
+	return NewRawVersion(versionRawString)
 }
 
 type PluginVersion struct {
--- a/version/version.go
+++ b/version/version.go
@@ -6,7 +6,6 @@
 package version
 
 import (
-	"bytes"
 	"fmt"
 
 	"github.com/hashicorp/go-version"
@@ -81,6 +80,7 @@ func NewPluginVersion(vers, versionPrerelease, versionMetadata string) *PluginVe
 	if versionMetadata != "" {
 		versionRawString = fmt.Sprintf("%s+%s", versionRawString, versionMetadata)
 	}
+
 	// This call initializes the SemVer to make sure that if Packer crashes due
 	// to an invalid SemVer it's at the very beginning of the Packer run.
 	semVer := version.Must(version.NewVersion(versionRawString))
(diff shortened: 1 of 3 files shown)


NorseGaud:
Looks good to me!

torarnv:
Nice! 👏🏻

version/version.go
```diff
@@ -55,22 +86,34 @@ type PluginVersion struct {
...
-	fmt.Fprintf(&versionString, "%s", p.version)
-	if p.versionPrerelease != "" {
-		fmt.Fprintf(&versionString, "-%s", p.versionPrerelease)
+func (p *PluginVersion) SetMetadata(meta string) {
+	p.versionMetadata = meta
+}
 
-		if GitCommit != "" {
-			fmt.Fprintf(&versionString, " (%s)", GitCommit)
+func (p *PluginVersion) FormattedVersion() string {
```
sylviamoss, line 102: Could this be a breaking change?

version/version.go
```diff
@@ -23,29 +22,61 @@ var Version = "0.5.3"
...
 // semantic versioning crash immediately rather than later.
+//
+// Deprecated: InitializePluginVersion does not support metadata out of the
+// box, and should be replaced by either NewRawVersion or NewPluginVersion.
 func InitializePluginVersion(vers, versionPrerelease string) *PluginVersion {
-	if vers == "" {
+	return NewPluginVersion(vers, versionPrerelease, "")
+}
+
+func NewRawVersion(rawSemVer string) *PluginVersion {
```
sylviamoss, line 48: Does it make sense to add comments to these two methods with examples of when to use them?

version/version.go
lbajolet-hashicorp: I don't know if I should call that a "breaking" change, but there's some potential extra information here, normally if a plugin creates their version with core + pre-release, the format should not change at all.
If there's a metadata however, this could change (metadata right now could be defined as `GitCommit` right now already, but it's global)

version/version.go
lbajolet-hashicorp: Oh yeah good call, I forgot to document them, I'll add this thanks!

version/version.go
sylviamoss: My concern was if Packer was using this same method and showing `git commit` between parenthesis it could be a breaking change for users checking git commits of builds, but Packer doesn't do that. Could it be a problem for packer plugin maintainers? I could see this happening if someone is using the plugin describe command for some validation or using the method to write user-agent to HTTP calls 🤔

lbajolet-hashicorp: Yeah that's a good callout, I didn't pay attention that I was changing the format honestly, I thought the commit was added as a metadata to the version if defined. For clarity though, it's not automagic, it needs either to be set by the plugin developers, or as a ldflag. AFAICT we don't set it in the scaffolding, or in any of our plugins.
Also, we use `String` for printing out the version of the plugin for `describe`. I'm not certain this is used anywhere, but I cannot certify it.

That said, I'll do a reroll of this commit to avoid changing the commit format, so it remains complete as it was beforehand.
I'll add a test to make sure it doesn't change between the two versions.

version/version_test.go
```diff
@@ -0,0 +1,121 @@
...
+			ver := NewPluginVersion(tt.coreVersion, tt.preVersion, tt.metaVersion)
+			verStr := ver.String()
+			if verStr != tt.expectVersionString {
+				t.Errorf("string format mismatch, version created is %q, expected %q", verStr, tt.expectVersionString)
+			}
+		})
+	}
+}
+
+func TestFormattedVersionString(t *testing.T) {
```
sylviamoss, line 101: Nice!

sylviamoss (approved):
LGTM! Thanks for working on the feedback 🎉

nywilken (approved):
Oh nice add! This looks good to me.

PR #229, opened 2024-03-22
[Feature Branch] Support protobuf/msgpack serialization for cty Values

lbajolet-hashicorp:
This PR aims to be a PoC/base for later developments on the wire-protocol for all the components that plugin expose.

We introduce a protobuf-serialisable data structure for `hcldec.HCLSpec`, which is used by all components of a plugin for describing the schema of what the plugin uses, and offers Packer a type to use to decode cty values from HCL code.

In doing so, we also explored using alternative serialisation formats for datasources, which were the only components leveraging cty's gob serialisation capabilities, and replace them with JSON/msgpack.

These changes make the SDK capable to communicate over-the-wire on recent hcl/cty versions, without needing to use the fork/replace for gob support.


plugin/set.go
```diff
@@ -158,13 +188,14 @@ func (i *Set) start(kind, name string) error {
...
-		Provisioners:   i.provisionersDescription(),
-		Datasources:    i.datasourceDescription(),
+		Version:         i.version,
+		SDKVersion:      i.sdkVersion,
+		APIVersion:      i.apiVersion,
+		Builders:        i.buildersDescription(),
+		PostProcessors:  i.postProcessorsDescription(),
+		Provisioners:    i.provisionersDescription(),
+		Datasources:     i.datasourceDescription(),
+		ProtocolVersion: "v2",
```
nywilken, line 198: We should add a constant for "v2" so that we can comment on what this value means.

plugin/set.go
```diff
@@ -106,11 +108,39 @@ func (i *Set) Run() error {
...
+
+	return append(args[:protobufPos], args[protobufPos+1:]...)
+}
+
 func (i *Set) RunCommand(args ...string) error {
 	if len(args) < 1 {
 		return fmt.Errorf("needs at least one argument")
 	}
 
+	args = useProtobuf(args...)
```
nywilken, line 142: Any particular reason why you chose the flag parsing logic above as opposed to adding a flag.Parse(args) when the issued sub command is "start".  

Then inside of the Start function when you construct the server you can specify that it should use protobuf for serialization. You will need to add a new filed to the server to track this. It can be a bool if you go the UseProtoBuf name or a string if you go ProtcolVersion. This way in the common rpc code the server can check if the field is set instead of the global `rpc.UseProto` variable.

rpc/common.go
```diff
@@ -7,12 +7,21 @@ import (
...
 	"github.com/hashicorp/hcl/v2/hcldec"
 	"github.com/zclconf/go-cty/cty"
+	"google.golang.org/protobuf/proto"
 )
 
+// UseProto lets us determine whether or not we should use protobuf for serialising
+// data over RPC instead of gob.
+//
+// This is controlled by Packer using the `--use-proto` flag on plugin commands.
+var UseProto bool = false
```
nywilken, line 23: I believe we should store this information on the server and client as fields. As opposed to using this global variable. 
In Packer when we check if the plugins support protocol_version "v2" we can config the client using PluginClientConfig to UseProto. The server as I mentioned above when passed the `--protobuf` flag will create the server with the UseProto field set to true.  


The code for the client and server in common then start to look like 

```go
 func (p *commonClient) ConfigSpec() hcldec.ObjectSpec {
	// Legacy: this will need to be removed when we discontinue gob-encoding
	//
	// This is required for backwards compatibility for now, but using
	// gob to encode the spec objects will fail against the upstream cty
	// library, since they removed support for it.
	//
	// This will be a breaking change, as older plugins won't be able to
	// communicate with Packer any longer.
	if !c.UseProto {  // or if c.ProtocolVersion != "v2"
		log.Printf("[DEBUG] - common: receiving ConfigSpec as gob")
		res := hcldec.ObjectSpec{}
		err := gob.NewDecoder(bytes.NewReader(resp.ConfigSpec)).Decode(&res)
		if err != nil {
			panic(fmt.Errorf("failed to decode HCL spec from gob: %s", err))
		}
		return res
	}
}

....

func (s *commonServer) ConfigSpec(_ interface{}, reply *ConfigSpecResponse) error {
// Legacy: this will need to be removed when we discontinue gob-encoding
	//
	// This is required for backwards compatibility for now, but using
	// gob to encode the spec objects will fail against the upstream cty
	// library, since they removed support for it.
	//
	// This will be a breaking change, as older plugins won't be able to
	// communicate with Packer any longer.
	if !s.UseProto {  // or if s.ProtocolVersion != "v2"
		log.Printf("[DEBUG] - common: receiving ConfigSpec as gob")
		res := hcldec.ObjectSpec{}
		err := gob.NewDecoder(bytes.NewReader(resp.ConfigSpec)).Decode(&res)
		if err != nil {
			panic(fmt.Errorf("failed to decode HCL spec from gob: %s", err))
		}
		return res
	}
}

```

I have to dive deeper into the datasources but I believe it embeds a Common client and a  Common server if so I believe the fields may be accessible on the datasource.

Makefile
```diff
@@ -59,7 +59,7 @@ generate: install-gen-deps ## Generate dynamically generated code
 	@find ./ -type f | xargs grep -l '^// Code generated' | xargs rm -f
 	PROJECT_ROOT="$(CURDIR)" go generate ./...
 	go fmt bootcommand/boot_command.go
-# 	go run ./cmd/generate-fixer-deprecations
+	protoc rpc/hcl_spec.proto --go_out=. --go_opt=paths=source_relative
```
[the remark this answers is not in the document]
lbajolet-hashicorp: No objection for this, if the tool works similarly we'll be good to go with it

plugin/server.go
```diff
@@ -54,7 +54,7 @@ const (
 	//
 	// Api version 4 did not have the notion of a minor version, so Packer will
 	// error with a weird error message.
-	APIVersionMajor, APIVersionMinor = "5", "0"
+	APIVersionMajor, APIVersionMinor = "5", "1"
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Old proto, this doesn't apply anymore

rpc/common.go
```diff
@@ -7,12 +7,30 @@ import (
...
 	"net/rpc"
+	"os"
+	"reflect"
 
 	"github.com/hashicorp/hcl/v2/hcldec"
 	"github.com/zclconf/go-cty/cty"
+	"google.golang.org/protobuf/proto"
 )
 
+const PackerUsePBEnvVar = "PACKER_RPC_PB"
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Old proto, this doesn't apply anymore

plugin/set.go
lbajolet-hashicorp: So `flag` I thought of using it yes, but in the end I went with this approach as the arguments are passed, and `flag.Parse` works on the raw arguments from `os.Args`, besides I think the order had some impact on what was returned, so I thought might as well do something ad-hoc that'd let us place the option wherever.
Besides that's not something that will be used outside Packer, and that'll possibly disappear someday (when Gob is out of the picture), so I thought it was maybe easier to do it this way rather than change how we manage flags in there today.

That being said there are merits to that option of refactoring to handle flags properly, just thought that today was not the day for that, but it should be feasible. Let me know if you think this'd be valuable and I can probably spend some time on that option.

nywilken: Ah I always confuse these args given the whole Run, RunCommand methods our CLI uses. Thanks for the details on why you wrote it this way. I don't think it is important at this very moment given that there is only one flag. But you wrote a decent snippet of code so I was curious why. Sorry for not looking at Run() too much.  

When ready I guess we can add the flag.Parse to the Run method and just create a variable for the UseProto flag.

rpc/common.go
nywilken: > I'm not sure as to the proposed change though, both clients and servers need to be in line with the protocol used, the current approach means UseProto can be global, it however needs to be set to either true or false ASAP in the lifecycle of the processes, that means at command start for plugins, and once the discovery is done for Packer.

This is correct. The caller would be responsible for initializing the server and client with the protocol enabled so since Packer is the control here. When we identify that all plugins support protocol v2 we setup Packer to start create a client for the plugin with the protocol set to use protobuf. The execute command that starts the plugin with the --protobuf flag. This will tell the SDK that the server should use protobuf. At this point the client and server will be in line. 


> E.g. from a plugin perspective, I'm pretty certain we won't have access to the command-flags at the point where we create the RPC structs, so we'll probably need to either forward it, or keep it globally. If we keep it global, are there merits to this approach, as we'll copy a global flag into a structure?

plugin/set.go
```diff
@@ -106,11 +116,39 @@ func (i *Set) Run() error {
 	return i.RunCommand(args...)
 }
 
+func (i *Set) useProtobuf(args ...string) []string {
```
nywilken, line 119: Just for our sake of our future self lets just document that this will fail for older plugins and plugins that don't support this flag.

plugin/set.go
```diff
@@ -131,6 +169,10 @@ func (i *Set) start(kind, name string) error {
 		return err
 	}
 
+	if i.useProto {
+		server.UseProto = true
+	}
+
```
nywilken, line 175:
```suggestion
	server.UseProto = i.useProto
```

plugin/set.go
```diff
@@ -115,11 +116,42 @@ func (i *Set) Run() error {
...
+	i.useProto = true
+
+	if protobufPos == 0 {
+		return args[1:]
+	}
+
+	if protobufPos == len(args)-1 {
+		return args[:len(args)-1]
+	}
```
nywilken, line 143:
```suggestion
```

Given that the upper index bound for a slice is cap(args) I believe we can drop these two if statements, as the final return will handle both cases. 

```go
//protobufPos == 0
return append(args[:0], args[1:])
//protobufPos == len(args)-1 (last item in the slice)
return append(args[:protobufPos], args[protobufPos:protobufPos) will return everything up to the flag and append an empty slice since it is the last item in the slice. 
```
I'll write a test to validate this.

nywilken (commented):
Overall this looks good. I left one small suggestion but I will add a test for it. As discussed  internally I am leaving a comment to mark the approach as approved but this is a feature branch that we will keep open and iterate on.

mogrogan (dismissed):
Left a question (although I know you probably have a good reason) but solid PR

LGTM

rpc/hcl_spec.proto
```diff
@@ -0,0 +1,94 @@
...
+*/
+message HCL2Spec {
+  map<string, Spec> TypeSpec = 1;
+}
+
+// A Spec is any kind of object that can convert losslessly to any of the hcldec.Spec types.
+message Spec {
+  oneof block {
+    Object object = 1;
+    // buf:lint:ignore FIELD_LOWER_SNAKE_CASE
```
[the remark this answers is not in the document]
lbajolet-hashicorp: That's a good question, I don't remember why that is; I'll try to change case here and we'll see what happens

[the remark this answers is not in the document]
lbajolet-hashicorp: To answer: this was lifted as-is from Nomad, but doesn't need to be. I rerolled the PR with `attr` lower-case, without the lint ignore directive as it's not necessary.
Tested locally with a qemu build, with both packer and the plugin linked to this version of the SDK, works seamlessly, so we can proceed with this instead.

mogrogan (approved):
LGTM

PR #232, opened 2024-04-16
Release 0.5.3

lbajolet-hashicorp:


nywilken (approved):
LGTM - after the release lets add the upgrade notes section so folks know how to update. 

```
Upgrading to this release may fail until you've applied one of the fixes documented in https://github.com/hashicorp/packer-plugin-sdk/issues/187#user-content-available-fixes. Consumers of the Packer plugin SDK require a replace directive within their plugin's go module file to point to a compatible version of go-cty. 

The replace directive subject to change in future releases can be applied by running the packer-sdc fix sub-command to apply the replace directive to your plugin with a recommended version of the go-cty fork.

Plugins already working with Packer Plugin SDK v0.5.2 are advised to apply the updated SDK fixes by re-running `packer-sdc fix` against the plugin's root directory. 
```

PR #243, opened 2024-06-03
communicator: fix reference to ssh_method in docs

lbajolet-hashicorp:
As pointed out in [a Packer PR](https://github.com/hashicorp/packer/pull/12949), the option's name does not match its real name for the ssh file transfer method.

This commit fixes that mismatch.


nywilken (approved):
I believe we need to run make Jen to update the plug-ins and possible Packer docs that are referencing this partial

lbajolet-hashicorp:
Yeah we'll need to update the dependency and `make generate`.
I suggest we do this once we've addressed the `gob` problem.

PR #247, opened 2024-06-13
Bump vault api dependancy to mitigate go-jose concerns, removes support for go1.20

JenGoldstrich:
I ran `go get github.com/hashicorp/vault/api` to mitigate concerns from Azure Plugin consumers with go-jose vulns, I understand these are not impacting the SDK directly but its worth an easy fix

go-jose v4 supports go 1.21 as the minimum version.


nywilken (approved):
LGTM

PR #257, opened 2024-07-10
Ability to change the HTTP network protocol

ejdre-vestas:
## Motivation

While trying to run the virtual-box-iso builder on a WSL2 environment, I had some issues getting the virtual box image to reach the HTTP server started by packer build. 
The VBox would stay stuck like this:

image

Eventually,I found that even though I was using the default "0.0.0.0" address, the server was listening only on IPv6 address (https://pkg.go.dev/net#Listen):

image
image

Because of this, the Windows host will only enable the loopback address on IPv6:
image

## Changes

This pull-request just makes it possible to specify the listening protocol with the property `http_network_protocol` which allows users to set the value to `tcp4` and overcome this issue.

## Note

These changes were tested by building a new binary from https://github.com/hashicorp/packer-plugin-virtualbox/pull/134 and referencing these SDK changes by adding a replace directive on the `go.mod` file.


multistep/commonsteps/http_config.go
```diff
@@ -58,6 +58,8 @@ type HTTPConfig struct {
 	// interface with a non-loopback address. Either `http_bind_address` or
 	// `http_interface` can be specified.
 	HTTPInterface string `mapstructure:"http_interface" undocumented:"true"`
+	// If true the HTTP server will only be bound to an IPv4 interface
+	HTTPOnlyIPv4 bool `mapstructure:"http_only_ipv4"`
```
lbajolet-hashicorp, line 62: If we're going with this, I would suggest allowing to specify the protocol here instead of adding a flag, this way we can repurpose it for tcp6/unix or any other protocol later if there's a need for that.

Something like:

```go
HTTPNetworkProtocol string `mapstructure:"http_network_protocol"`
```

Which could be specified as any network type that `Listen` accepts (`Prepare` would then be responsible for ensuring the value is valid).

lbajolet-hashicorp (changes requested):
Hi @ejdre-vestas,

Thanks for the PR!

I've left a comment on the approach, while it works, I think it's too reductive to have a flag for IPv4 only, as I can imagine the counterpoint of having the HTTP server IPv6 only, hence my suggestion to add an enumeration instead of a boolean flag for this purpose.

That being said, I'm surprised regarding the behaviour you're experiencing. My understanding is that by default the server listening on `tcp` means that IPv4 and IPv6 are both supported, which is the behaviour I experience locally (Ubuntu 22.04), but I imagine that can fluctuate between systems, and therefore the option makes sense.

Feel free to address my suggestion when you can, and I'll come back for another pass of review then.

Thanks again!

multistep/commonsteps/http_config.go
ejdre-vestas: Hi @lbajolet-hashicorp,

Sorry, I've been offline. Thanks a lot for looking at these pull-requests. I have pushed a new commit with the proposed changes.

PR #272, opened 2024-12-18
commonsteps: StepDownload timeout override

mogrogan:
Should now be overridable via the environment variable `PACKER_GETTER_READ_TIMEOUT`

### DIFF
--- a/multistep/commonsteps/step_download.go
+++ b/multistep/commonsteps/step_download.go
@@ -64,40 +64,59 @@ type StepDownload struct {
 // The timeout must be long enough to accommodate large/slow downloads.
 const defaultGetterReadTimeout time.Duration = 30 * time.Minute
 
-var defaultGetterClient = getter.Client{
-	// Disable writing and reading through symlinks.
-	DisableSymlinks: true,
-	// The order of the Getters in the list may affect the result
-	// depending if the Request.Src is detected as valid by multiple getters
-	Getters: []getter.Getter{
-		&getter.GitGetter{
-			Timeout: defaultGetterReadTimeout,
-			Detectors: []getter.Detector{
-				new(getter.GitHubDetector),
-				new(getter.GitDetector),
-				new(getter.BitBucketDetector),
-				new(getter.GitLabDetector),
+var getterReadTimeout = defaultGetterReadTimeout
+var defaultGetterClient = getter.Client{}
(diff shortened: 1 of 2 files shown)


multistep/commonsteps/step_download.go
```diff
@@ -64,40 +64,57 @@ type StepDownload struct {
...
+var getterReadTimeout = defaultGetterReadTimeout
+var defaultGetterClient = getter.Client{}
+
+func init() {
+	prepareGetterClient()
+}
+
+func prepareGetterClient() {
+	getterReadTimeout := defaultGetterReadTimeout
+	if env, exists := os.LookupEnv("PACKER_GETTER_READ_TIMEOUT"); exists {
```
lbajolet-hashicorp, line 76: Note to myself: we'll need to add that new envvar to the docs somewhere

multistep/commonsteps/step_download_test.go
```diff
@@ -287,6 +289,47 @@ func TestStepDownload_download(t *testing.T) {
...
 
+func TestStepDownload_short_timeout(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Log("skipping download test on windows right now.")
+		return
+	}
+
+	os.Setenv("PACKER_GETTER_READ_TIMEOUT", "1ns")
+	prepareGetterClient()
+	defer prepareGetterClient()
```
lbajolet-hashicorp, line 300: ~~That seems weird, since you invoke it synchronously before~~
^ As I was writing this comment, I realised this was to re-initialise the client with the default values, without `PACKER_GETTER_READ_TIMEOUT`, I assume this is also why you don't use `t.Setenv` for setting the environment variable in the context of this test?
This seems OK to me if that's the rationale, but I would very much consider adding a comment here to explain that reasoning as it is not super straightforward when coming back to this code.

Note: as with the other comment I left, this can be addressed separately in a future PR, no need to leave it out of the SDK for now.

lbajolet-hashicorp (dismissed):
Hey @mogrogan,

Overall LGTM, thanks. I left a few comments on the code, which are more nitpicks/docs changes than anything else. Feel free to let me know how you want to proceed with those, and we might be able to merge this today.

lbajolet-hashicorp (approved):

PR #281, opened 2025-03-17
fixes for packer_acc_tests for windows 

kp2099:
Changing the path here to reference it to the bootstrap_win.txt file thats been added here (https://github.com/hashicorp/packer/pull/13334)

Currently i believe the acceptance tests fail with "user_data_file" not found.

Also adding skip_create_ami to not create an AMI when running acceptance tests


anurag5sh (approved):
LGTM!

PR #283, opened 2025-03-22
windows sshagent forwarding

anshulsharma-hashicorp:
Issue: https://github.com/hashicorp/packer/issues/3670

I am adding the changes to handle the ssh-agent forwarding in windows, before this changes the packer code would always go and check the SSH_AUTH_SOCK environment variable for socket file, which is the behavior in linux but in windows a named static files gets used for this purpose, so have added changes for the same.

This was reported by multiple users for reference i have added the github issue link above.

Tests: 
Have added the Unit Test cases for the new changes. 

Closes https://github.com/hashicorp/packer/issues/3670


sdk-internals/communicator/ssh/windows_named_pipe.go
```diff
@@ -0,0 +1,27 @@
...
+package ssh
+
+import (
+	"fmt"
+	"github.com/Microsoft/go-winio"
+	"net"
+)
+
+func GetSSHAgentConnection() (net.Conn, error) {
+	pipePath := "\\\\.\\pipe\\openssh-ssh-agent"
```
JenGoldstrich, line 19: Where does this pipe path come from?

JenGoldstrich (commented):
Thanks for this change @anshulsharma-hashicorp, I think it's looking good, although I do wonder if handling the inclusion of this logic using a go:build directive is best, I would defer to @lbajolet-hashicorp's opinions for code style there though.

Can you provide me some more context on where that Windows pipe path comes from, and can you also sign the CLA agreement

sdk-internals/communicator/ssh/windows_named_pipe.go
anshulsharma-hashicorp: this is a static named pipe in windows, which holds the ssh private key for ssh-agent forwarding.

anurag5sh: Can we declare this pipe name as constant on top of the file? It would give that some context as to a magic string.

kp2099:
The change looks harmless to me, although i don't have enough expertise on windows named pipes.

sdk-internals/communicator/ssh/windows_named_pipe.go
JenGoldstrich: Agree, it gives me a bit of pause that this is jsut a magic string, Lucas and I did some digging in docs and couldn't find any direct reference to this, only like stack overflow pages, so I worry that someone else reading this code in the future wouldn't have any context on what this is, I think making it a constant + adding a comment would be wise here

anshulsharma-hashicorp: Thanks for the suggestion i have done this.

anurag5sh (commented):
Changes look good to me. Additionally we could add any reference to the named pipe path in windows if any.

sdk-internals/communicator/ssh/windows_named_pipe.go
anshulsharma-hashicorp: here adding the log of testing on windows machine https://docs.google.com/document/d/1wVG_NntN2OpCcF2QPUeJiA6WndSvEaoE-VTOY5xKHrM/edit?tab=t.0

JenGoldstrich (dismissed):
LGTM! Thank you for this contribution @anshulsharma-hashicorp I know you've put a lot of work into this, you did a good job understanding the full SSH stack here, and helping improve my knowledge.  🚀 Ship it!

lbajolet-hashicorp (dismissed):

lbajolet-hashicorp (approved):

JenGoldstrich (approved):

PR #290, opened 2025-04-24
Deprecating `Message` method in UI interface

tanmay-hc:
Deprecating the `Message` method from the cli UI interface to simplify the logs and replacing them with the Say method.


anshulsharma-hashicorp (approved):
LGTM

PR #293, opened 2025-05-29
Add bootcommand mappings for left/right command and option keys

torarnv:
When automating macOS systems it's convenient to refer to the keys by their natural names, and for the VNC driver none of the existing special keys mapped to option, so this enables boot commands using the option key.

PR #323, opened 2026-04-02
[WIP] feat: add support for meta keys

PrashantRaj18198:
### Description
Added support for meta keys which map to 'Option' keys on macs. We can use `<leftMetaOn/Off>` and `<rightMetaOn/Off>` to signal the left and right Option key press.

I haven't checked if the keypress maps to something in linux, let me know if this is required.

### Rollback Plan

If a change needs to be reverted, we will roll out an update to the code within 7 days.

NA


tanmay-hc:
Closing this PR as this was fixed with https://github.com/hashicorp/packer-plugin-sdk/pull/293

PR #340, opened 2026-07-10
Sort secrets longest-first when redacting

apoorva-01:
### Description

Redaction ranged over a map, so the replace order was random. When one sensitive value is a substring of another (`ubuntu` and `ubuntu-22.04` both sensitive), the short one could get replaced first and leave the tail of the longer secret in the log:

```
connecting to ubuntu-22.04 now  ->  connecting to <sensitive>-22.04 now
```

Same line came out either way run to run. Sorting longest-first before replacing fixes it. Tests for `FilterString` and `Write` fail on the old code and pass with this.

Found while checking at hashicorp/packer#13659. cc @tanmay-hc

### Resolved Issues

None.

### Rollback Plan

If a change needs to be reverted, we will roll out an update to the code within 7 days.

Yes, log redaction. Makes it stricter: overlapping secrets now redact the longest match first, so less can slip through.

### DIFF
--- a/packer/logs.go
+++ b/packer/logs.go
@@ -6,6 +6,7 @@ package packer
 import (
 	"bytes"
 	"io"
+	"sort"
 	"strings"
 	"sync"
 )
@@ -31,10 +32,8 @@ func (l *secretFilter) SetOutput(output io.Writer) {
 }
 
 func (l *secretFilter) Write(p []byte) (n int, err error) {
-	for s := range l.s {
-		if s != "" {
-			p = bytes.Replace(p, []byte(s), []byte("<sensitive>"), -1)
-		}
+	for _, s := range l.secrets() {
+		p = bytes.ReplaceAll(p, []byte(s), []byte("<sensitive>"))
 	}
 	return l.w.Write(p)
 }
@@ -42,12 +41,32 @@ func (l *secretFilter) Write(p []byte) (n int, err error) {
 // FilterString will overwrite any senstitive variables in a string, returning
 // the filtered string.
 func (l *secretFilter) FilterString(message string) string {
-	for s := range l.s {
-		if s != "" {
-			message = strings.Replace(message, s, "<sensitive>", -1)
-		}
+	for _, s := range l.secrets() {
+		message = strings.ReplaceAll(message, s, "<sensitive>")
 	}
 	return message
+}
+
+// secrets returns the non-empty registered secrets ordered longest first.
+// Ranging over the map directly is randomly ordered, so when one secret is a
+// substring of another (e.g. "ubuntu" and "ubuntu-22.04") the shorter one could
+// be replaced first, leaving the remainder of the longer secret in the output.
+// Redacting the longest match first keeps that from happening and makes the
+// result deterministic.
+func (l *secretFilter) secrets() []string {
+	secrets := make([]string, 0, len(l.s))
+	for s := range l.s {
+		if s != "" {
+			secrets = append(secrets, s)
+		}
+	}
+	sort.Slice(secrets, func(i, j int) bool {
+		if len(secrets[i]) != len(secrets[j]) {
+			return len(secrets[i]) > len(secrets[j])
+		}
+		return secrets[i] < secrets[j]
+	})
+	return secrets
 }
 
 var LogSecretFilter secretFilter
--- a/packer/logs.go
+++ b/packer/logs.go
@@ -4,7 +4,6 @@
 package packer
 
 import (
-	"bytes"
 	"io"
 	"sort"
 	"strings"
@@ -12,9 +11,10 @@ import (
 )
 
 type secretFilter struct {
-	s map[string]struct{}
-	m sync.Mutex
-	w io.Writer
+	s      map[string]struct{}
+	sorted []string // non-empty secrets, longest first; rebuilt on Set
+	m      sync.Mutex
+	w      io.Writer
 }
 
 func (l *secretFilter) Set(secrets ...string) {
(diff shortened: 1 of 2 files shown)


tanmay-hc:
While this is not related to the https://github.com/hashicorp/packer/issues/13659 issue, this is still a good find.
Thanks for the PR.

packer/logs.go
```diff
@@ -31,23 +32,41 @@ func (l *secretFilter) SetOutput(output io.Writer) {
...
+		p = bytes.ReplaceAll(p, []byte(s), []byte("<sensitive>"))
 	}
 	return l.w.Write(p)
 }
 
 // FilterString will overwrite any senstitive variables in a string, returning
 // the filtered string.
 func (l *secretFilter) FilterString(message string) string {
+	for _, s := range l.secrets() {
+		message = strings.ReplaceAll(message, s, "<sensitive>")
```
tanmay-hc: Replacements are applied to the result of previous replacements, so a later secret can match text introduced by the `<sensitive>` marker.
For example, with secrets `"my_token"` and `"sitive"`, longest-first changes "my_token" into `"<sensitive>"` and then into `"<sen<sensitive>>"`.
This PR makes that malformed output deterministic and may break consumers that recognize the exact redaction marker.
Please compute matches against the original input, or otherwise ensure generated markers cannot be filtered again, and cover both FilterString and Write with a regression test.

packer/logs.go
```diff
@@ -31,23 +32,41 @@ func (l *secretFilter) SetOutput(output io.Writer) {
...
 		}
 	}
-	return message
+	sort.Slice(secrets, func(i, j int) bool {
+		if len(secrets[i]) != len(secrets[j]) {
+			return len(secrets[i]) > len(secrets[j])
+		}
+		return secrets[i] < secrets[j]
+	})
+	return secrets
```
tanmay-hc, line 106: Could we avoid allocating and sorting the complete secret set for every message? `FilterString` runs for every `BasicUi.Say` and `BasicUi.Error`, and `Write` may be installed in a logging path.
This changes each call from direct map iteration to a slice allocation plus an `O(NlogN)` sort.
Since secrets change only through Set, please maintain a sorted snapshot when Set mutates the set under the mutex, then let readers use that snapshot. This would also provide a race-free solution to the concurrent map access noted above.

packer/logs_test.go
```diff
@@ -0,0 +1,51 @@
...
+// When one registered secret is a substring of another, the shorter value must
+// not be replaced first. Ranging over the map is randomly ordered, so without
+// longest-first ordering the short secret ("ubuntu") can be redacted before the
+// long one ("ubuntu-22.04"), leaving the tail "-22.04" (part of a real secret)
+// in the output. The loop makes the random map order show up reliably.
+func TestSecretFilterFilterStringOverlapping(t *testing.T) {
+	const in = "connecting to ubuntu-22.04 now"
+	const want = "connecting to <sensitive> now"
+
+	for i := 0; i < 100; i++ {
```
tanmay-hc: This test relies on unspecified map iteration eventually selecting the shorter secret first within 100 attempts. That is probabilistic and is not guaranteed to fail against the old implementation on every Go version or platform. Please assert the sorted snapshot/order directly, while retaining one end-to-end redaction assertion.

tanmay-hc (changes requested):
Refer to the copilot's Race condition comments too.

apoorva-01:
Reworked it, covers all five:

**Races** — `Set` now builds the longest-first list under the lock, and reads grab that slice and the writer together under the lock (new `snapshot()`). Nothing iterates the map or reads `w` unlocked anymore.

**Marker re-matching** — `redact` makes one pass over the original input instead of chaining `ReplaceAll`, so a `<sensitive>` marker it writes is never searched again. `my_token` + `sitive` stays `<sensitive>` instead of `<sen<sensitive>>`.

**Per-message sort** — the sort happens in `Set`, not per message. `FilterString`/`Write` reuse the stored snapshot, so there's no allocation or sort on the hot path.

**Test** — asserts the sorted order directly now, plus deterministic redaction cases and a `-race` test for concurrent Set/read, instead of leaning on map iteration landing the wrong order within 100 tries.

Passes `go test -race ./packer/`.

packer/logs.go
```diff
@@ -31,23 +33,77 @@ func (l *secretFilter) SetOutput(output io.Writer) {
...
 
 func (l *secretFilter) Write(p []byte) (n int, err error) {
-	for s := range l.s {
-		if s != "" {
-			p = bytes.Replace(p, []byte(s), []byte("<sensitive>"), -1)
-		}
-	}
-	return l.w.Write(p)
+	secrets, w := l.snapshot()
+	return w.Write([]byte(redact(string(p), secrets)))
```
tanmay-hc, line 37: The `Write` returns the byte count from the downstream writer, but after redaction the string can be longer than the original p (secrets expand to <sensitive>). This means callers will see a short-write on every redaction even when nothing went wrong.
Can we rewrite the function to something like:
```
func (l *secretFilter) Write(p []byte) (n int, err error) {
    secrets, w := l.snapshot()
    _, err = w.Write([]byte(redact(string(p), secrets)))
    if err == nil {
        n = len(p)
    }
    return
}
```

packer/logs.go
```diff
@@ -31,23 +33,77 @@ func (l *secretFilter) SetOutput(output io.Writer) {
...
+		}
+	}
+	return b.String()
+}
+
+// sortedSecrets returns the non-empty secrets ordered longest first, and
+// lexicographically for equal lengths so the order is deterministic. Redacting
+// the longest match first keeps a shorter secret that is a substring of a
+// longer one (e.g. "ubuntu" and "ubuntu-22.04") from leaving the remainder of
+// the longer secret in the output.
```
tanmay-hc, line 92: The comments on these functions seem a bit too big unnecessarily. Probably due to AI iterations. Can you shorten them to be more precise?

tanmay-hc (changes requested):
Changes mostly look good just please do a rebase and address the two minor comments and we can merge the PR.

tanmay-hc:
@apoorva-01 Ping.

PR #344, opened 2026-07-24
refactor: remove redundant else branches

tenthirtyam:
### Description

Removes unnecessary `else` clauses after terminating statements.

### Resolved Issues

Redundant `else` clauses in `if` statements that contain a `return`.

### Rollback Plan

Revert commit.

None.

### DIFF
--- a/acctest/pluginacc.go
+++ b/acctest/pluginacc.go
@@ -116,9 +116,9 @@ func TestPlugin(t *testing.T, testCase *PluginTestCase) {
 					"acceptance test template can be found at %s",
 					err.Error(), filepath.Join(cwd, initLogfile),
 					filepath.Join(cwd, templatePath))
-			} else {
-				os.Remove(initLogfile)
 			}
+
+			os.Remove(initLogfile)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


acctest/pluginacc.go
```diff
@@ -156,8 +156,8 @@ func TestPlugin(t *testing.T, testCase *PluginTestCase) {
 			"acceptance test template can be found at %s",
 			checkErr.Error(), filepath.Join(cwd, logfile),
 			filepath.Join(cwd, templatePath))
-	} else {
-		os.Remove(templatePath)
-		os.Remove(logfile)
 	}
+
+	os.Remove(templatePath)
+	os.Remove(logfile)
```
tanmay-hc, line 162: I feel we would like to keep the two temporary files in case of test failures as they could help the users debug the failed test better.

Same goes for the change in `acctest/provisioneracc/provisioners.go`

tenthirtyam: Good catch on the intent to keep the template and log on failure for debugging (the `Fatalf` message points at those paths for that reason).

Behaviorally this is unchanged: `t.Fatalf` never returns (`FailNow` / `runtime.Goexit`), so the `os.Remove `calls after the if only run on success. Same for p`rovisioners.go`. Dropping the `else` after a terminating statement is the usual Go early-exit idiom.

Happy to restore the `else` in both places if you prefer that form to make the “cleanup only on success” intent more obvious at a glance.

tenthirtyam:
@tanmay-hc - rebased and ready for review. Please see my review comment reply at https://github.com/hashicorp/packer-plugin-sdk/pull/344#discussion_r3708375507. Thank you!

PR #346, opened 2026-07-24
refactor: use reflect.TypeFor for type comparisons

tenthirtyam:
### Description

Replace runtime sample values with direct generic type lookup.

### Resolved Issues

In Go 1.22 and later, `reflect.TypeFor[T]()` returns the same `reflect.Type` without requiring a runtime value.

### Rollback Plan

Revert commit.

None.

### DIFF
--- a/template/config/decode.go
+++ b/template/config/decode.go
@@ -310,10 +310,9 @@ func stringToTrilean(f reflect.Type, t reflect.Type, v interface{}) (interface{}
 	// *bool values in order to intelligently default, even when the values are
 	// being set by a template variable.
 
-	testTril, _ := TrileanFromString("")
-	if t == reflect.TypeOf(testTril) {
+	if t == reflect.TypeFor[Trilean]() {
 		// From value is string
-		if f == reflect.TypeOf("") {
+		if f == reflect.TypeFor[string]() {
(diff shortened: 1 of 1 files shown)


template/config/decode.go
```diff
@@ -310,18 +310,18 @@ func stringToTrilean(f reflect.Type, t reflect.Type, v interface{}) (interface{}
 	// *bool values in order to intelligently default, even when the values are
 	// being set by a template variable.
 
-	testTril, _ := TrileanFromString("")
-	if t == reflect.TypeOf(testTril) {
+	_, _ = TrileanFromString("")
```
tanmay-hc: nit: I don't think we need this line now. can we remove this?

tenthirtyam: Definitely seems unused. I'll update this one.

tenthirtyam: Resolved in https://github.com/hashicorp/packer-plugin-sdk/commit/286248234995d49ddaab37375945e5f7becb989d. [Comparison Link](https://github.com/hashicorp/packer-plugin-sdk/compare/b531b68e306e55f38e8b7b20b2cac6cfc0d51bd8..286248234995d49ddaab37375945e5f7becb989d)

tanmay-hc (approved):

PR #350, opened 2026-07-24
refactor: simplify scp mode detection

tenthirtyam:
### Description

Simplifies `bytes.IndexByte` using `bytes.Contains`.

### Resolved Issues

Use direct byte membership checks because only flag presence matters, making the intent clearer and avoiding an unused index.

### Rollback Plan

Revert commit.

None.

### DIFF
--- a/adapter/adapter.go
+++ b/adapter/adapter.go
@@ -218,11 +218,11 @@ func (c *Adapter) scpExec(args string, in io.Reader, out io.Writer) error {
 	}
 	rest = strings.Join(shargs, "")
 
-	if i := bytes.IndexByte(opts, 't'); i >= 0 {
+	if bytes.IndexByte(opts, 't') >= 0 {
 		return scpUploadSession(opts, rest, in, out, c.comm)
 	}
 
-	if i := bytes.IndexByte(opts, 'f'); i >= 0 {
+	if bytes.IndexByte(opts, 'f') >= 0 {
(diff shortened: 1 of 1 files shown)


adapter/adapter.go
```diff
@@ -218,11 +218,11 @@ func (c *Adapter) scpExec(args string, in io.Reader, out io.Writer) error {
 	}
 	rest = strings.Join(shargs, "")
 
-	if i := bytes.IndexByte(opts, 't'); i >= 0 {
+	if found := bytes.Contains(opts, []byte{'t'}); found {
 		return scpUploadSession(opts, rest, in, out, c.comm)
```
tanmay-hc, line 222: If the goal is to just remove the unused index, we can simply use the following code instead:
```
if bytes.IndexByte(opts, 't') >= 0 {
```
Using `bytes.Contains` here forces an inline slice allocation `([]byte{'t'})`, which is less idiomatic and visually noisier than a simple byte literal.

tenthirtyam: Addressed in https://github.com/hashicorp/packer-plugin-sdk/commit/e88cc1d7f72c0ee74bbf221f4632a2c9b6f3a262. [Comparison Link](https://github.com/hashicorp/packer-plugin-sdk/compare/afe8c1ef1cdd778f7076d6cd950a4bc338df3145..e88cc1d7f72c0ee74bbf221f4632a2c9b6f3a262)

tanmay-hc (approved):

PR #334, opened 2026-06-23
feat: add winrm connect timeout and retry params

Pandapip1:
Submitted on behalf of a third-party: [Realtime Robotics](https://rtr.ai)

### Description

WinRM (at least on tiny11) seems to have an internal 70-second timeout when it hasn't started yet that limits the rate at which WinRM connections are retried. This adds a timeout so that it can fail and retry early, to more quickly respond to when WinRM _does_ become available.

Untested, but does compile and seems trivial enough.

### Resolved Issues

*N.A.*

### Rollback Plan

*N.A.*

### Changes to Security Controls

*N.A.*

PR #354, opened 2026-08-03
feat: add winrm connect timeout and retry params

tanmay-hc:
WinRM (at least on tiny11) seems to have an internal 70-second timeout when it hasn't started yet that limits the rate at which WinRM connections are retried. This adds a timeout so that it can fail and retry early, to more quickly respond to when WinRM does become available.

Original PR: #334


Pandapip1 (approved):
Thanks, this does still look correct

anurag5sh (approved):

PR #356, opened 2026-08-03
refactor: drop ineffective omitempty

tenthirtyam:
### Description

`encoding/json` treats struct values as never empty, so `json:",omitempty"` on the embedded `OnlyExcept` field in `PostProcessor` and `Provisioner` has no effect.

Removes the misleading tags; `OnlyExcept `already applies `omitempty` to its `Only` and `Except` slice fields, and JSON output is unchanged.

### Resolved Issues

`encoding/json` treats struct values as never empty, so `json:",omitempty"` on the embedded `OnlyExcept` field in `PostProcessor` and `Provisioner` has no effect.

### Rollback Plan

Revery commit.

None.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -104,7 +104,7 @@ func (b *Builder) MarshalJSON() ([]byte, error) {
 
 // PostProcessor represents a post-processor within the template.
 type PostProcessor struct {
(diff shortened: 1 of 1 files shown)


tenthirtyam:
@tanmay-hc - rebased and ready for review.

